PR #6, opened 2026-04-21
Codex/fix ssh key loading

asmuelle:


GOODBOY008:
@asmuelle I think your patch is better than my fix. I'm willing to see your contribution. Open source for fun. 👍

PR #17, opened 2026-05-28
fix: show terminal scrollbar

### LINKED ISSUE #14: Terminal panel has no visible vertical scrollbar for long output
The terminal panel does not show a vertical scrollbar on the right side when the output exceeds the visible area. This makes it hard to know the current scroll position or quickly scroll back through long command output. Mouse wheel scrolling may work, but a visible scrollbar is expected for a terminal UI, especially for long outputs such as logs, help messages, build output, or benchmark results. Please consider enabling the xterm.js viewport scrollbar or adding a setting to show/hide the terminal scrollbar.

htazq:
## Root cause

`PtyTerminal` used the `terminal-no-scrollbar` container class and injected CSS that hid `.xterm-viewport` scrollbars in WebKit, Firefox, and legacy Edge/IE. xterm could still scroll, but the vertical scrollbar was intentionally invisible for long terminal output.

## Fix

- Replaced the hidden-scrollbar container class with `pty-terminal-container`.
- Removed `display: none`, `scrollbar-width: none`, and `-ms-overflow-style: none` from the xterm viewport styles.
- Added explicit thin scrollbar styling for the xterm viewport while preserving the existing transparent-background terminal selectors.
- Added a Vitest regression test that fails if hidden-scrollbar CSS comes back.

Closes #14

## Verification

RED/GREEN:
- RED confirmed: `pnpm.cmd test src/__tests__/pty-terminal-scrollbar.test.tsx` failed before the fix because `.terminal-no-scrollbar` and hidden scrollbar CSS were present.
- GREEN confirmed: `pnpm.cmd test src/__tests__/pty-terminal-scrollbar.test.tsx` passed after the fix.

Passed:
- `pnpm.cmd test src/__tests__/pty-terminal-scrollbar.test.tsx`
- `pnpm.cmd test src/__tests__/pty-terminal-activation.test.tsx`
- `pnpm.cmd exec tsc --noEmit`
- `pnpm.cmd lint` (0 errors; existing warnings remain)
- `pnpm.cmd build`

Known local failures unrelated to this change:
- `pnpm.cmd test` still fails in existing suites: `src/__tests__/connection.test.ts` needs a Tauri runtime/invoke bridge, `src/__tests__/file-entry-types.test.ts` has Windows path separator expectation failures, and `src/__tests__/terminal-group-serializer.test.ts` fails when saving a fixture without `tabToGroupMap`. The new scrollbar test passes in the full run.
- `cd src-tauri && cargo test` fails locally because `openssl-sys` cannot find `perl` while building vendored OpenSSL.

### DIFF
--- a/src/__tests__/pty-terminal-scrollbar.test.tsx
+++ b/src/__tests__/pty-terminal-scrollbar.test.tsx
@@ -1,10 +1,11 @@
 import React from 'react';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
-import { cleanup, render } from '@testing-library/react';
+import { act, cleanup, render } from '@testing-library/react';
 import { PtyTerminal } from '../components/pty-terminal';
 
 const mocks = vi.hoisted(() => {
   const terminals: Array<any> = [];
+  let capturedLineFeedCallback: (() => void) | null = null;
 
   class MockTerminal {
     cols = 80;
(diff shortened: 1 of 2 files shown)


GOODBOY008:
@htazq I add patch commit to enhance this pr.

GOODBOY008 (approved):

PR #18, opened 2026-05-28
fix: support terminal tab drag reorder

### LINKED ISSUE #16: Feature request: Support drag-and-drop reordering for terminal tabs
It would be helpful if terminal tabs could be reordered by drag and drop. Currently, after opening multiple SSH sessions, the tab order appears to be fixed, and users cannot freely drag an active tab to another position. When working with many servers or grouping related sessions, being able to rearrange tabs manually would make the workflow much more convenient. Please consider adding drag-and-drop tab reordering, similar to common terminal clients, browsers, or tools like MobaXterm.

htazq:
## Root cause

The tab drag handlers were attached only to the inner tab list whose width is the rendered tabs. When a user dragged a tab to the empty area at the end of the tab strip, no `drop` event reached the reorder handler, so the tab could appear impossible to move to the end.

## Fix

- Moved the tab drag-over, drag-leave, and drop handlers to the outer tab bar so the whole strip is a drop target.
- Let the inner tab list grow with `flex-1 min-w-0`, preserving overflow scrolling while exposing the empty strip area for drops.
- Added a component-level Vitest regression test that drags the active tab into the empty strip area and expects `REORDER_TAB` to move it to the end.

Closes #16

## Verification

RED/GREEN:
- RED confirmed: `pnpm.cmd test src/__tests__/group-tab-bar-drag-reorder.test.tsx` failed before the fix with 0 dispatch calls.
- GREEN confirmed: `pnpm.cmd test src/__tests__/group-tab-bar-drag-reorder.test.tsx` passed after the fix.

Passed:
- `pnpm.cmd test src/__tests__/group-tab-bar-drag-reorder.test.tsx`
- `pnpm.cmd test src/__tests__/terminal-group-reducer.test.ts src/__tests__/terminal-group-reducer.property.test.ts`
- `pnpm.cmd exec tsc --noEmit`
- `pnpm.cmd lint` (0 errors; existing warnings remain)
- `pnpm.cmd build`

Known local failures unrelated to this change:
- `pnpm.cmd test` still fails in existing suites: `src/__tests__/connection.test.ts` needs a Tauri runtime/invoke bridge, `src/__tests__/file-entry-types.test.ts` has Windows path separator expectation failures, and `src/__tests__/terminal-group-serializer.test.ts` fails when saving a fixture without `tabToGroupMap`. The new drag reorder test passes in the full run.
- `cd src-tauri && cargo test` was checked during this issue pass and fails locally because `openssl-sys` cannot find `perl` while building vendored OpenSSL.

### DIFF
--- a/src/__tests__/group-tab-bar-drag-reorder.test.tsx
+++ b/src/__tests__/group-tab-bar-drag-reorder.test.tsx
@@ -1,0 +1,130 @@
+import React from 'react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { fireEvent, render } from '@testing-library/react';
+import { GroupTabBar } from '../components/terminal/group-tab-bar';
+import type { TerminalTab } from '../lib/terminal-group-types';
+
+const mocks = vi.hoisted(() => ({
+  dispatch: vi.fn(),
+}));
+
+vi.mock('../lib/terminal-group-context', () => ({
+  useTerminalGroups: () => ({
+    dispatch: mocks.dispatch,
+  }),
+}));
+
+vi.mock('../components/ui/button', () => ({
+  Button: ({
+    children,
+    variant: _variant,
+    size: _size,
+    ...props
+  }: React.ButtonHTMLAttributes<HTMLButtonElement> & {
+    variant?: string;
+    size?: string;
+  }) => <button type="button" {...props}>{children}</button>,
+}));
+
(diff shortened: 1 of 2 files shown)


GOODBOY008:
@htazq I have tested on my local with your branch but it does not worked.

GOODBOY008:
impled in main commit：397e8cbb8dd77fa95f7d8d9b0cb3ab080ccc1427

PR #34, opened 2026-06-10
fix: preserve terminal scrollback history

htazq:
﻿## Problem
Long terminal output cannot be scrolled back far enough. After commands like `seq 1 20000`, early output becomes unreachable and the terminal appears to stop around a few hundred lines of history.

## Root cause
Commit `95bd715` lowered the default xterm.js scrollback from `10000` to `500` to reduce memory usage. xterm's visible buffer is roughly `scrollback + viewport rows`, so a 500-line scrollback with a ~90-row terminal shows up as about 590 reachable lines. Saved `terminalAppearance.scrollback` values from that regression could also keep overriding future defaults.

## Fix
- Restore the default xterm scrollback to `10000` lines.
- Add scrollback normalization constants: min `1000`, default `10000`, max `50000`.
- Migrate invalid or too-small saved scrollback values, including the regressed `500`, back to the default.
- Clamp saved settings so users can raise scrollback up to `50000` without unbounded growth.
- Apply `term.options.scrollback` in-place when terminal appearance settings change, without recreating or disposing the terminal.
- Add a regression test using real xterm.js that writes `20000` lines and verifies the default buffer keeps at least `10000` lines.
- Update stale terminal test mocks for the current `PtyTerminal` dependencies and paste behavior.

## Validation
- PASS: `pnpm test -- src/__tests__/terminal-scrollback.test.ts`
- PASS: `pnpm test -- src/__tests__/terminal-scrollback.test.ts src/__tests__/pty-terminal-scrollbar.test.tsx src/__tests__/pty-terminal-activation.test.tsx`
- PASS: `pnpm test -- src/__tests__/terminal-scrollback.test.ts src/__tests__/terminal-group-serializer.test.ts`
- PASS: `pnpm lint` (0 errors, existing warnings remain)
- PASS: `pnpm build`
- PARTIAL: `pnpm test` runs 425/432 tests successfully, but `src/__tests__/connection.test.ts` still fails because it calls Tauri `invoke` in jsdom where `window.__TAURI_INTERNALS__` is unavailable. That failure is unrelated to this change and exists outside the terminal scrollback path.
- PARTIAL: `cargo test` was attempted from `src-tauri` but exceeded the local 184s timeout before producing a final result.

## Risk
This increases the default in-memory terminal history from 500 to 10000 lines, so per-terminal memory usage can increase. The history remains bounded by xterm.js' scrollback ring buffer; this is not infinite DOM appending and does not render output into ordinary divs. User-configured scrollback is capped at `50000` lines. Existing PTY/WebSocket output flow-control and output-limit behavior is left in place.

## Manual test
```sh
yes "scrollback-test" | head -n 20000
seq 1 20000
```

Expected:
- With default settings, output completion should allow scrolling back through at least the most recent 10000 lines.
- With scrollback configured to 50000, the full 20000-line manual test output should be reachable.
- Switching tabs and resizing the browser/app window should not clear the terminal buffer.

### DIFF
--- a/src/__tests__/pty-terminal-activation.test.tsx
+++ b/src/__tests__/pty-terminal-activation.test.tsx
@@ -34,6 +34,7 @@ const mocks = vi.hoisted(() => {
     getSelection = vi.fn(() => '');
     selectAll = vi.fn();
     clear = vi.fn();
+    reset = vi.fn();
     dispose = vi.fn();
   }
 
@@ -88,7 +89,7 @@ vi.mock('@xterm/addon-web-links', () => ({
 
 vi.mock('@xterm/addon-webgl', () => ({
   WebglAddon: vi.fn(function WebglAddon() {
-    return { dispose: vi.fn() };
+    return { dispose: vi.fn(), onContextLoss: vi.fn() };
   }),
 }));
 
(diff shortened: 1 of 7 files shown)


GOODBOY008 (approved):
@htazq LGTM

PR #44, opened 2026-06-22
feat(i18n): integrate react-i18next for app internationalization

GOODBOY008:


GOODBOY008:
@htazq Would you help me to test this ?

htazq:
> @htazq Would you help me to test this ?

Sure, I’d be happy to help test it when I have some time. I’ll take a look and let you know if I find any issues.

htazq:
Thanks! I tested this PR locally and it works well on my side.

I verified the i18n switching between English and Simplified Chinese, and also tested it in an active SSH session. The UI updates correctly, the settings page looks good, and I didn’t notice any obvious issues during normal usage.

One small note: the current language options are Auto, English, and Simplified Chinese only. That looks fine for this PR since it mainly introduces the i18n framework and the initial translations.

Nice work! 
image

image

PR #47, opened 2026-07-16
fix(terminal): preserve live sessions when moving tabs between groups

### LINKED ISSUE #46: Moving a tab to a new group remounts terminals and restarts PTY sessions
## Summary

Moving an open terminal tab to a new group disconnects and recreates every live terminal in the original group.

This is not only a visual refresh: the existing WebSocket and backend PTY sessions are closed, so foreground commands, interactive programs, terminal history, and unsaved terminal state can be lost.

## Steps to reproduce

1. Open multiple SSH terminal tabs in one group.
2. Start an interactive or long-running command in one or more tabs.
3. Right-click a tab.
4. Select **Move to New Group** and choose any direction.
5. Observe the terminal startup sequence again:

```text
Starting interactive shell (WebSocket + PTY mode)...
WebSocket connected
PTY connection started
```

## Actual behavior

All terminals in the source group are unmounted. Their cleanup closes the WebSocket and PTY, then the new layout mounts fresh terminal instances and starts new shells.

## Expected behavior

Moving or splitting tabs should only change the layout. Existing terminal components, WebSocket connections, PTY sessions, history, and running commands should remain intact.

## Root cause

`MOVE_TAB_TO_NEW_GROUP` replaces a grid leaf with a branch. At that React tree position, the rendered type changes from `TerminalGroupView` to a `ResizablePanelGroup`. React therefore unmounts the old `TerminalGroupView`, which currently owns every `PtyTerminal` in that group.

Stable keys alone cannot retain components that move across different parents or whose parent hierarchy changes.

The issue was observed on Windows, but the lifecycle path is platform-independent.

## Related

- #7 reports focus/render restoration problems during ordinary tab switching. It is related to terminal component lifecycle, but it has a different trigger and does not specifically cover the WebSocket/PTY teardown caused by moving a tab to a new group.

htazq:
## Summary

- keep live tab content mounted in stable per-tab portal containers
- let the recursive grid move lightweight portal hosts instead of remounting terminals
- preserve active-group behavior, active-tab visibility, reconnect behavior, and normal cleanup when a tab is actually closed
- add a regression test that moves a live terminal tab into a new group without mounting or unmounting either terminal again

## Root cause

`MOVE_TAB_TO_NEW_GROUP` replaces a grid leaf with a branch. React consequently unmounts the old `TerminalGroupView`, which previously owned every live `PtyTerminal` in the source group. The cleanup closes each WebSocket and backend PTY, and the replacement layout starts new shells.

Stable keys cannot preserve components that move across different parents or whose parent hierarchy changes, so this separates live tab ownership from the layout tree.

## User impact

Moving or splitting tabs now changes only their layout. Existing xterm instances, WebSocket connections, PTY sessions, terminal history, and running foreground commands stay alive.

## Testing

- `pnpm test -- src/__tests__/terminal-tab-portals.test.tsx src/__tests__/grid-renderer.test.tsx` ? 2 files, 5 tests passed
- `pnpm build` ? passed
- scoped ESLint for the three changed terminal components ? passed with one pre-existing `grid-renderer.tsx` warning
- full Vitest run ? 31 files / 467 tests passed; 14 pre-existing environment/mock failures remain in `connection.test.ts`, `pty-terminal-activation.test.tsx`, and `pty-terminal-scrollbar.test.tsx`
- full ESLint remains blocked by the pre-existing `settings-modal.tsx:293` error

Fixes #46
Related to #7

### DIFF
--- a/src/__tests__/terminal-tab-portals.test.tsx
+++ b/src/__tests__/terminal-tab-portals.test.tsx
@@ -1,0 +1,167 @@
+import React from 'react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { cleanup, render, screen } from '@testing-library/react';
+import type { TerminalGroupState } from '../lib/terminal-group-types';
+import {
+  TerminalTabPortalHost,
+  TerminalTabPortalProvider,
+} from '../components/terminal/terminal-tab-portals';
+
+const lifecycle = vi.hoisted(() => ({
+  mounted: vi.fn(),
+  unmounted: vi.fn(),
+  dispatch: vi.fn(),
+}));
+
+let mockState: TerminalGroupState;
+
+vi.mock('../lib/terminal-group-context', () => ({
+  useTerminalGroups: () => ({
+    state: mockState,
+    dispatch: lifecycle.dispatch,
+  }),
+}));
+
+vi.mock('../lib/terminal-callbacks-context', () => ({
+  useTerminalCallbacks: () => ({}),
+}));
+
+vi.mock('../components/pty-terminal', async () => {
+  const ReactModule = await import('react');
+
+  return {
+    PtyTerminal: ({ connectionId }: { connectionId: string }) => {
+      ReactModule.useEffect(() => {
+        lifecycle.mounted(connectionId);
+        return () => lifecycle.unmounted(connectionId);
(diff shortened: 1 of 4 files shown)


GOODBOY008 (approved):
@htazq LGTM

PR #52, opened 2026-07-17
feat(terminal): press R to reconnect disconnected sessions

### LINKED ISSUE #12: Feature request: allow pressing `R` to reconnect from a disconnected terminal
## Feature request

When an SSH session is disconnected or offline, it would be useful to support a keyboard-based reconnect shortcut, similar to MobaXterm: press `R` in the disconnected terminal panel to reconnect directly.

## Current behavior

When the connection is lost, the terminal shows a message similar to:

```text
[Connection failed permanently. Use right-click → Reconnect to retry.]
```

At this point, the user needs to use the mouse, right-click the terminal/session, and choose `Reconnect` manually.

## Expected behavior

When the active terminal/session is in a disconnected state:

- Pressing `R` or `r` reconnects the current session directly.
- The shortcut should only be active when the session is disconnected/offline.
- It should not affect normal connected terminal input.
- It should work when the disconnected terminal tab/panel is focused.

Optional UX improvement:

```text
Connection failed permanently. Press R to reconnect, or right-click → Reconnect.
```

## Why this is useful

For SSH management tools, reconnect is a high-frequency operation, especially when servers reboot, network drops, or the SSH daemon restarts.

A one-key reconnect flow is faster than opening the context menu every time and matches the behavior many users are familiar with from MobaXterm.

## Suggested implementation idea

Reuse the existing reconnect action behind the context menu. Add a keydown handler for the disconnected terminal state only:

```ts
if (isDisconnectedSession(activeSession) && event.key.toLowerCase() === 'r') {
  reconnectSession(activeSession.id);
}
```

The key handler should be scoped carefully so it does not intercept normal shell input while the SSH session is connected.

htazq:
## Summary
- allow `R` / `r` to reconnect only the active disconnected terminal
- keep connected terminal input unchanged and ignore modified, repeated, or IME key events
- honor the existing Auto Reconnect setting when a PTY connection closes
- show localized manual reconnect guidance after automatic reconnect is disabled or exhausted

## Verification
- `pnpm test src/__tests__/pty-terminal-activation.test.tsx src/__tests__/terminal-tab-portals.test.tsx src/__tests__/keyboard-shortcuts-hook.test.tsx src/__tests__/keyboard-shortcuts.test.ts` (40 passed)
- `pnpm exec tsc --noEmit`
- `pnpm exec eslint src/components/pty-terminal.tsx src/components/terminal/terminal-tab-portals.tsx --quiet`
- `pnpm i18n:check`
- `pnpm build`

Full `pnpm test`: 476 passed; the 7 existing `src/__tests__/connection.test.ts` cases still require a Tauri runtime and fail in plain Vitest because `window.__TAURI_INTERNALS__.invoke` is unavailable.

Closes #12

### DIFF
--- a/src/__tests__/pty-terminal-activation.test.tsx
+++ b/src/__tests__/pty-terminal-activation.test.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { act, cleanup, render } from '@testing-library/react';
 import { PtyTerminal } from '../components/pty-terminal';
+import { APP_SETTINGS_STORAGE_KEY } from '../lib/keyboard-shortcuts';
 
 const mocks = vi.hoisted(() => {
   const terminals: Array<any> = [];
@@ -201,6 +202,7 @@ describe('PtyTerminal activation', () => {
     mocks.terminals.length = 0;
     mocks.fitAddons.length = 0;
     mocks.webSockets.length = 0;
+    localStorage.removeItem(APP_SETTINGS_STORAGE_KEY);
 
     Object.defineProperty(HTMLElement.prototype, 'offsetWidth', {
       configurable: true,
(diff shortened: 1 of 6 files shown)


GOODBOY008 (approved):
Thanks @htazq for this nice quality-of-life feature — pressing R on a disconnected terminal is a great shortcut, and the guard conditions (active tab only, no modifiers, IME-safe via `isComposing`/`keyCode === 229`, no repeat) are exactly right.

To get it mergeable against the current main (the terminal subsystem moved quite a bit since you opened this), I rebased your branch onto latest `main` (now including #153's lazy session restore) and force-pushed to your fork. Resolution notes:

- **`pty-terminal.tsx`**: kept your functional code (`isAutoReconnectEnabled()` helper, the `autoReconnect === false` early-return branch with `connectionClosedManualReconnect`, `APP_SETTINGS_STORAGE_KEY` import). For the failure messages I kept main's existing keys (`autoReconnectFailed` with `{{attempts}}`, `reconnectFailedPermanently`) and merged your "Press R to reconnect" wording into them, instead of the duplicate `{{count}}`/`connectionFailedPermanently` keys.
- **`terminal-tab-portals.tsx`**: your `handleReconnectKeyDown` kept alongside main's newer `handleDetach`; the `onKeyDown` prop wires up cleanly.
- **Tests**: kept both sides — main's newer activation tests plus your "stops after a disconnect when automatic reconnect is disabled" and the tab-portals R-key routing test.
- **Locales**: dropped the duplicate `autoReconnectFailed`/`connectionFailedPermanently` keys (wording merged into main's), kept your `connectionClosedManualReconnect`, and added the missing Polish translation for it.

All gates green locally: `tsc --noEmit` clean, `i18n:check` 1184 keys × 3 locales, eslint 0 errors, 813/813 vitest tests pass. Thanks again for the contribution!

PR #61, opened 2026-07-24
fix(sidebar): fix folder click not selecting and right-click expanding unrelated nodes

sunxiaobin89:
## Summary

Fix two issues with folder node interaction in the connection sidebar tree:

1. **Clicking a folder row would not select it** — it only triggered expand/collapse.
2. **Right-clicking a folder would not select it first** before showing the context menu, unlike connection nodes.
3. **Clicking or right-clicking a folder could unexpectedly expand other folders** due to unstable component state references causing full tree rebuilds.

## Root Cause

All three issues stem from the same root: `App.tsx::handleConnectionSelect` filtered out nodes with `type === "folder"`, silently ignoring folder selection. Combined with unstable `activeConnections` prop references (a new `Set` created on every render), this triggered `ConnectionManager` tree rebuilds that reset `isExpanded` to `true` for all folders.

### `src/App.tsx`
- Remove folder type filter in `handleConnectionSelect` — folders can now be selected
- Memoize `activeConnections` with `useMemo` to prevent unnecessary tree rebuilds

### `src/components/connection-manager.tsx`
- Move `toggleExpanded` from row click handler (`handleNodeClick`) to the chevron button — row click now only selects, chevron click controls expand/collapse
- Preserve `isExpanded` state during tree rebuild via `mergeExpanded` to prevent folder selection from collapsing or expanding unrelated directories

### Files changed
- `src/App.tsx` — 2 changes
- `src/components/connection-manager.tsx` — 2 changes
- `src/__tests__/connection-manager-folder-selection.test.tsx` — new file, 10 tests
- `.gitignore` — add `test-results/`

## Testing

- **504 tests pass** (36 test files, 0 failures, 0 regressions)
- **10 new tests** covering:
  - Folder click selects the folder node
  - Folder click does not collapse children
  - Chevron click toggles expand/collapse
  - Chevron click does not trigger row selection (stopPropagation)
  - Visual highlight (`bg-accent`) applied to selected folders
  - No standalone `bg-accent` class when another node is selected
  - Connection click still selects (regression)
  - Folders preserve collapsed state after tree rebuild
  - Folders preserve expanded state after tree rebuild
- Manual testing on macOS: behavior confirmed correct

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ dist-ssr
 *.local
 target
 .qoder
+test-results/
 
 # Editor directories and files
 .vscode/*
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -95,6 +95,13 @@ function AppContent() {
     return Object.values(state.groups).flatMap(g => g.tabs);
   }, [state.groups]);
 
+  // Memoized set of active connection IDs — stable reference prevents
+  // ConnectionManager from rebuilding its tree on every parent render.
+  const activeConnectionIds = useMemo(
+    () => new Set(allTabs.map(tab => tab.id)),
+    [allTabs],
+  );
+
   // Apply stored language preference (follows OS locale when set to "auto")
   useEffect(() => {
     void applyLanguageFromPreference();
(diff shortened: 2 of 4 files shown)


GOODBOY008 (approved):
LGTM

PR #62, opened 2026-07-24
fix(connection-dialog): prevent number input default value override on clear

sunxiaobin89:
## Summary

When clearing a number input field in the connection dialog (port, proxy port, etc.), the value immediately snaps back to its default (22, 8080, etc.) instead of allowing the user to type a new value.

## Root Cause

In  handlers using the pattern `parseInt(e.target.value) || 22`, clearing the field produces `NaN || 22 = 22`, which `updateConfig()` immediately writes back — React controlled input then renders `22`.

## Changes

- **src/components/connection-dialog.tsx** — Added a unified `displayValues` state object (`number | ''`) that tracks what each number input should visually display, separate from `ConnectionConfig`'s strict `number` type. Introduced a generic `handleNumberInput()` helper to handle all 4 affected inputs (port, proxyPort, keepAliveInterval, serverAliveCountMax) with the same logic.

## Testing

- 494 tests pass, 0 new failures
- Manually verified in real app: clearing port now stays empty, typing a valid number works, protocol switching resets port to default

### DIFF
--- a/src/components/connection-dialog.tsx
+++ b/src/components/connection-dialog.tsx
@@ -97,6 +97,39 @@ export function ConnectionDialog({
 
   const [config, setConfig] = useState<ConnectionConfig>(defaultConfig);
 
+  // Track number input display values separately from config to allow
+  // the field to be empty while editing — React controlled inputs need
+  // value="" to render empty, but ConnectionConfig uses strict number types.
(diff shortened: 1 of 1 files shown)


GOODBOY008 (approved):
@sunxiaobin89 Great catch! LGTM~

PR #27, opened 2026-06-03
fix: bound PTY websocket output buffering

htazq:
## Summary

- Bound the PTY -> WebSocket output queue and add timeout/drop behavior so slow frontend/WebSocket consumers cannot grow backend memory without bound.
- Send PTY output as binary WebSocket frames instead of JSON `Vec<u8>` payloads, while keeping the existing JSON path as a compatibility fallback.
- Make Pause/Resume actually pause the PTY reader and clean up active PTY sessions on WebSocket disconnect.

Closes #26

## Testing

- `cargo test websocket_server::tests`
- `cargo check`
- `cargo test --lib -- --skip ssh::tests::key_loading_tests::test_connect_missing_key_file_returns_error`
- `npm exec --yes pnpm@9.15.4 -- exec tsc --noEmit`
- `npm exec --yes pnpm@9.15.4 -- test src\__tests__\pty-terminal-activation.test.tsx src\__tests__\pty-terminal-scrollbar.test.tsx`
- `npm exec --yes pnpm@9.15.4 -- lint`
- `rustfmt --edition 2021 --check src\websocket_server.rs`
- `git diff --check`

Note: full `cargo test` currently has one unrelated environment-sensitive failure: `ssh::tests::key_loading_tests::test_connect_missing_key_file_returns_error` attempts to connect to `127.0.0.1:22` and gets connection refused before the missing-key assertion. `cargo fmt --check` also reports unrelated pre-existing formatting in `src-tauri/src/commands.rs`.

Issue #26, opened 2026-06-03
Windows: r-shell.exe virtual memory grows unbounded during high PTY output


htazq:
## Summary

On Windows, `r-shell.exe` can grow virtual memory / commit usage continuously under sustained terminal output until the system reports low virtual memory.

## Observed evidence

Windows event evidence from an affected run:

```text
2026/6/2 22:14:54
r-shell.exe PID 2580 used 30,239,735,808 bytes

2026/6/2 22:19:58
r-shell.exe PID 2580 used 32,418,369,536 bytes

2026/6/2 22:26:24
r-shell.exe PID 2580 used 34,773,286,912 bytes
```

The same period showed OneDrive around 1.3 GB and Edge around 0.8 GB, so they were not the primary contributor. Windows later reported low virtual memory and unrelated services/processes failed to allocate memory.

## Likely root cause

The PTY output path can buffer without an effective upper bound when the frontend/WebSocket sender is slower than the remote PTY output rate:

```text
PTY reader -> accumulated Vec<u8> -> clone -> JSON String -> unbounded mpsc queue -> WebSocket sender -> frontend/xterm
```

In `src-tauri/src/websocket_server.rs`, PTY output was serialized as JSON `WsMessage::Output { data: Vec<u8> }` and sent through an unbounded Tokio channel. `Pause` / `Resume` messages were acknowledged but did not actually pause the PTY reader, so slow consumption could keep accumulating backend memory.

## Reproduction ideas

Run a high-output command in a PTY session and monitor `r-shell.exe` on Windows:

```bash
yes
while true; do seq 1 100000; done
journalctl -f
tail -f /var/log/syslog
```

PowerShell monitor:

```powershell
while ($true) {
  Get-Process r-shell | Select-Object Id,
    @{n='WS_MB';e={[math]::Round($_.WorkingSet64/1MB,1)}},
    @{n='Private_MB';e={[math]::Round($_.PrivateMemorySize64/1MB,1)}},
    @{n='Virtual_MB';e={[math]::Round($_.VirtualMemorySize64/1MB,1)}}
  Start-Sleep 10
}
```

## Expected behavior

High PTY output, background tabs, frontend stalls, or slow WebSocket sends should not allow backend memory to grow linearly without bound. The backend should apply bounded queues, backpressure, drop/coalescing, and cleanup stale PTY tasks on disconnect.

GOODBOY008:
[...]

PR #66, opened 2026-07-27
fix: keep PTY output flow control bounded

### LINKED ISSUE #60: Bug: terminal clears all scrollback after only 2 MiB of cumulative output
## Summary

r-shell automatically resets and clears the terminal after only 2 MiB of
cumulative output:

[Output limit reached — scrollback cleared to free memory]

This occurs during normal operational workloads such as fio, iostat, top,
watch, build logs, and multi-job performance tests.

The SSH/WebSocket/PTY session remains alive, but all previous terminal history
is lost.

## Steps to reproduce

1. Open an SSH terminal.
2. Run a high-output but normal command, for example:

   fio --name=test --rw=write --bs=1m --numjobs=128 ...

   or:

   yes

3. Wait until approximately 2 MiB of output has been received.
4. Observe that r-shell resets and clears the terminal.

## Actual behavior

The implementation counts all cumulative decoded output:

sessionOutputRef.current += data.length

When it reaches 2 MiB, it executes:

term.reset();
term.clear();

This deletes scrollback and resets terminal state, even when xterm's actual
retained scrollback is already bounded.

Programs that redraw the same status line using carriage returns also count
toward the limit, despite not retaining equivalent amounts of scrollback.

## Expected behavior

- Normal high-output commands must not unexpectedly erase terminal history.
- The configured xterm scrollback limit should bound retained terminal memory.
- Output limits should be based on actual retained buffer usage, not cumulative
  session output.
- `term.reset()` should not be used for routine memory management.
- Any emergency memory protection threshold should be configurable.
- If trimming is necessary, discard only the oldest scrollback lines and
  preserve the current viewport and terminal modes.

## Suggested solution

1. Remove the cumulative 2 MiB session output limit.
2. Rely on xterm's bounded `scrollback` option.
3. Keep the existing RAF write batching and watermark flow control.
4. Add a configurable scrollback setting, for example:
   - 500
   - 2,000
   - 10,000
   - 50,000 lines
5. If an emergency memory cap is still required, measure actual buffer or
   memory usage and trim only old history without calling `term.reset()`.

## Environment

- r-shell current main build
- Windows
- Reproduced with fio using 128 jobs

htazq:
## Summary

- return exactly one PTY output credit for each WebSocket frame after xterm processes its RAF batch
- immediately return credits for frames buffered by the streaming UTF-8 decoder without producing text
- remove the cumulative 2 MiB `reset()` / `clear()` path and rely on the existing bounded xterm scrollback
- add regression coverage for initial credits, single and batched frames, split UTF-8, sustained output, and tab activation

## Root cause

The backend consumes one semaphore permit per output frame, but the frontend replenished a fixed four permits after processing any low-water batch. `creditsGranted` was never decremented when frames arrived and was reset immediately after granting, so a low-rate stream could grow the backend credit window by roughly three permits per frame.

The separate 2 MiB guard counted all decoded output since the last clear, not retained terminal memory. Normal sustained output eventually called both `term.reset()` and `term.clear()` even though xterm scrollback is already bounded by its configured line limit.

After this change, the invariant is: available credits plus frames awaiting xterm completion remain equal to the initial window. An incomplete UTF-8 sequence retains only the decoder's pending bytes and returns that frame credit immediately, avoiding deadlock.

## Related work

- Fixes #60.
- #34 already bounds configurable xterm scrollback; this change relies on that existing mechanism.
- #26 / #27 addressed the downstream WebSocket queue and binary frames.
- #30 targets possible russh-side accumulation when downstream reading stalls. It is related but distinct and is intentionally not included here.

## Validation

Reproduced on current `main` (`547e1bb`) before the fix:

- one processed frame: 6 `Resume` messages instead of 3 total
- three batched frames: 6 instead of 5 total
- split UTF-8 across two frames: 6 instead of 4 total
- sustained output over 2 MiB: `term.reset()` called once

Passing after the fix:

- `pnpm test -- src/__tests__/pty-terminal-activation.test.tsx src/__tests__/terminal-scrollback.test.ts src/__tests__/terminal-tab-portals.test.tsx` - 14 passed
- `pnpm test` - 569 passed
- `pnpm exec tsc --noEmit`
- `pnpm i18n:check` - 950 keys in both locales
- `pnpm build`
- `pnpm exec eslint src/components/pty-terminal.tsx --quiet`
- `git diff --check`

Repository-wide `pnpm lint` remains blocked by the pre-existing `src/components/settings-modal.tsx:308` `no-unnecessary-type-assertion` error; the changed source file passes strict lint. Local `cargo test` reached the unrelated `openssl-sys` build script and exceeded the 304-second validation limit; this PR changes no Rust files, and PR CI runs the Rust suite on Windows, macOS, and Ubuntu.

## Risk / rollback

The change reduces the maximum in-flight output to the existing two-frame window. If xterm stops completing writes, output backpressures instead of accumulating permits. Rollback is the single commit in this PR.

### DIFF
--- a/src/components/pty-terminal.tsx
+++ b/src/components/pty-terminal.tsx
@@ -38,13 +38,6 @@ interface PtyTerminalProps {
  * Communication is done via WebSocket for low-latency bidirectional streaming.
  */
 
-/** Per-session output cap. When cumulative bytes written to xterm exceed this
- *  value the scrollback is cleared automatically so V8 heap stays bounded.
- *  2 MB of decoded text ≈ ~25k typical 80-char terminal lines. Kept low to
- *  prevent V8 heap fragmentation and WebGL texture-cache bloat during
- *  sustained high-throughput output (e.g. `yes`). */
-const SESSION_OUTPUT_LIMIT_BYTES = 2 * 1024 * 1024;
-
 export function PtyTerminal({
   connectionId,
   connectionName,
@@ -100,8 +93,6 @@ export function PtyTerminal({
   const autoReconnectAfterDropRef = React.useRef(0);
   const MAX_AUTO_RECONNECT_AFTER_DROP = 5;
 
-  // Cumulative bytes written to xterm this session — reset on clear.
-  const sessionOutputRef = React.useRef(0);
   const inputEncoderRef = React.useRef(new TextEncoder());
 
   const sendInputToPty = React.useCallback((data: string): boolean => {
@@ -379,9 +370,8 @@ export function PtyTerminal({
 
     // RAF write batching state — lifted to effect scope so cleanup can cancel.
     let writeBuffer = '';
-    let watermark = 0;
+    let bufferedFrameCount = 0;
     let rafId: number | null = null;
-    let creditsGranted = 0;
     
     // CRITICAL: Wait for terminal to have proper dimensions before connecting
     // Hidden terminals (display: none) may have cols=10, rows=5 which breaks PTY
@@ -464,7 +454,7 @@ export function PtyTerminal({
       };
 
       // =========================================================================
-      // RAF-Based Write Batching + Watermark Flow Control
+      // RAF-Based Write Batching + Credit Flow Control
       //
       // Based on xterm.js best practices:
       // - http://xtermjs.org/docs/guides/flowcontrol/
@@ -478,21 +468,8 @@ export function PtyTerminal({
       // Solution:
       // 1. Accumulate all incoming frames in a string buffer.
       // 2. Flush once per requestAnimationFrame (~60 writes/s instead of 100+).
-      // 3. Use watermark-based flow control: send Resume credits only when the
-      //    pending byte count drops below LOW_WATER, avoiding per-frame ACKs.
+      // 3. Return exactly one credit per frame after xterm processes the batch.
       // =========================================================================
-
-      /** High watermark (bytes): above this, the buffer is considered "full" and
-       *  we stop granting credits until xterm drains below LOW_WATER.  128 KB
-       *  keeps the emulator snappy for keystrokes under fast input (xterm guide
-       *  recommends ≤ 500 KB for responsiveness). */
-      const HIGH_WATER = 128 * 1024;
-      /** Low watermark (bytes): below this, we grant a batch of credits to the
-       *  backend so it can send more data. */
-      const LOW_WATER = 16 * 1024;
-      /** How many credits to grant each time watermark drops below LOW_WATER.
-       *  Keeps the pipeline flowing without flooding the WS receive queue. */
-      const CREDIT_BATCH = 4;
 
       const grantCredits = (count: number) => {
         if (ws.readyState === WebSocket.OPEN) {
@@ -501,7 +478,6 @@ export function PtyTerminal({
           for (let i = 0; i < count; i++) {
             ws.send(msg);
           }
-          creditsGranted += count;
         }
       };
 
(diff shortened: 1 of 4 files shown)


htazq:
Refreshed this PR onto current `main` (`547e1bb`) and re-ran the regression loop. The four focused tests fail on `main` with the original credit counts / 2 MiB reset, then pass with this one-commit branch. Full frontend suite is 569/569; cross-platform CI is running on the refreshed head.

GOODBOY008 (commented):
Reviewed the rebased head (`45890f0`, which includes the maintainer rebase onto current `main` — conflict resolution in `pty-terminal.tsx` plus removal of the now-unused `ptyTerminal.outputLimitReached` locale keys).

**Verdict: correct, and the invariant finally holds end-to-end.** What I verified against the backend protocol:

1. **Exact credit accounting.** The backend reader consumes exactly one permit per sent frame (`credits.acquire()` + `forget()` before *each* `flush_output` — both branches in `websocket_server.rs`), and this change returns exactly one credit per received frame: `frameCount` is captured at flush time and granted only from the `term.write()` callback, i.e. after xterm has actually processed the batch. The old leak (`creditsGranted` zeroed right after every grant, so the `creditsGranted < CREDIT_BATCH * 2` guard never bound anything) is gone along with the variable.
2. **No deadlock paths.** `writeBuffer` and `bufferedFrameCount` are only mutated together synchronously; frames arriving mid-write reschedule the flush; and the empty-decode immediate grant covers the streaming decoder holding an incomplete UTF-8 sequence. I couldn't construct a sequence that loses a permit or double-grants.
3. **Bounded under stall.** If rAF stops (hidden window), credits stop returning, the backend blocks on `credits.acquire()`, and buffering is capped at ~2 frames + the 256-message WS queue + the 16 KB accumulator — a few MB worst case, self-recovering on visibility.
4. **2 MiB reset removal is safe.** Scrollback is hard-capped at `MAX_TERMINAL_SCROLLBACK = 50000` lines with xterm trimming internally, so the removed guard bought nothing except destroyed history/alt-screen state (#60).
5. **Tests are substantive** — they hold the write callback to prove grants happen only after xterm processes, assert exact permit totals (2→3, 2+3, 2+1+1 for split UTF-8, 132 across >2 MiB), and verify decoder reassembly across frames.

CI is green on all three platforms. Two non-blocking notes inline.

src/components/pty-terminal.tsx
```diff
@@ -601,8 +567,7 @@ export function PtyTerminal({
                 signalReady(connectionId);
                 // Credit-based flow control: seed the pipeline with initial
                 // credits so the PTY reader can start sending immediately.
-                // Ongoing credits are managed by the watermark-based flow
-                // control in the flush callback above.
+                // Ongoing credits are returned by the flush callback above.
                 const INITIAL_WINDOW = 2;
```
GOODBOY008, line 571: The window of 2 is adequate today, but only because of a coupling with the backend: the reader flushes at most `OUTPUT_FLUSH_BYTES = 16 KB` per `OUTPUT_FLUSH_INTERVAL_MS = 10`, capping production at ~1.6 MB/s — just below what 2 frames per rAF cycle (~2 MB/s at 60 Hz) can drain. If either backend constant is ever raised (e.g. for high-throughput `cat` of large files), this window silently becomes the bottleneck and would need to scale with it. Worth a one-line note here so future tuning doesn't miss it. Non-blocking.

src/components/pty-terminal.tsx
```diff
@@ -478,30 +468,16 @@ export function PtyTerminal({
...
-      const LOW_WATER = 16 * 1024;
-      /** How many credits to grant each time watermark drops below LOW_WATER.
-       *  Keeps the pipeline flowing without flooding the WS receive queue. */
-      const CREDIT_BATCH = 4;
-
       const grantCredits = (count: number) => {
         if (ws.readyState === WebSocket.OPEN) {
           // Send a single Resume per credit (backend Semaphore.add_permits(1))
           const msg = JSON.stringify({ type: 'Resume', connection_id: connectionId });
           for (let i = 0; i < count; i++) {
```
GOODBOY008, line 478: Each credit is its own WS message, so a burst of N frames flushed together (e.g. output queued while the window was hidden) sends N tiny JSON `Resume` frames back-to-back. Fine at current volumes, but a natural follow-up would be extending the protocol to `Resume { connection_id, count }` with the backend doing `add_permits(count)` — the frontend already knows the batch size here. Non-blocking.

src/components/pty-terminal.tsx
```diff
@@ -544,8 +503,15 @@ export function PtyTerminal({
       };
 
       const enqueueOutput = (text: string) => {
+        // A streaming decoder can retain an incomplete UTF-8 sequence without
+        // producing text. It has already consumed the frame, so return that
+        // credit immediately to avoid stalling on multi-frame characters.
+        if (!text) {
+          grantCredits(1);
```
GOODBOY008, line 510: Nice catch — without this immediate return, a multi-byte character split across frames would park a permit in the decoder forever (the frame is consumed but produces no text), eventually starving the backend reader after INITIAL_WINDOW such frames. The split-`中` regression test covers exactly this.

GOODBOY008 (approved):
Approving after full review (see previous COMMENTED review for the detailed verification): credit invariant holds end-to-end against the backend protocol, no permit-loss paths, bounded under rAF stall, 2 MiB reset removal safe under the 50k-line scrollback cap, substantive regression tests, CI green on all platforms. The two inline notes are non-blocking follow-ups (Resume batch count, INITIAL_WINDOW coupling note). CI: green.

PR #67, opened 2026-07-27
fix(edit-connection): improve tab sync, credential handling, and save flow

sunxiaobin89:
## Summary

Improves the edit-connection experience: tab titles now update after a rename, credentials are preserved when the edit dialog opens from a failed connection, and the "Update & Connect" button becomes "Save" to avoid unnecessary reconnection.

### Edit-connection flow
- **`UPDATE_TAB_NAME` reducer action** — tab title updates live after saving a renamed connection
- **`handleSave` replaces `handleConnect` in edit mode** — the button now says "Save" and doesn't trigger a reconnect
- **`pendingConnectionId` tracking** — when the edit dialog was opened from a failed double-click (e.g. wrong port), saving the fix auto-retries the connection
- **Existing tab reuse** — the auto-retry reuses the previously failed tab instead of opening a duplicate
- **6 `setEditingConnection` calls now include credentials** — `password`, `privateKeyPath`, and `passphrase` are no longer lost when a connection error opens the edit dialog

### New-connection / general
- **SSH config saved before connecting** — aligns with SFTP/FTP/Desktop behavior, so a config is preserved even if the server is temporarily unreachable
- **Unique session IDs for each SSH attempt** — prevents stale backend sessions from interfering with updated connection parameters
- **Sidebar refresh after dialog close** — `ConnectionManager` now refreshes its tree when the dialog closes, so saved changes appear immediately

### Stability
- **`PtyTerminal` effect no longer depends on `connectionName`** — renaming a connection no longer tears down the WebSocket + PTY session
- **`UPDATE_TAB_NAME` + `RECONNECT_TAB` reducer tests added**

## Testing

- 510 tests pass (37 test files)
- TypeScript compiles cleanly
- Tauri build succeeds

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -63,6 +63,11 @@ function AppContent() {
   const [connectionInitialFolder, setConnectionInitialFolder] = useState<string | undefined>();
   const [settingsModalOpen, setSettingsModalOpen] = useState(false);
   const [editingConnection, setEditingConnection] = useState<ConnectionConfig | null>(null);
+  // Track whether the edit dialog was opened due to a failed connection attempt (double-click)
+  // vs. direct edit (right-click). When non-null and matches saved config id, auto-connect after save.
+  const [pendingConnectionId, setPendingConnectionId] = useState<string | null>(null);
+  // Incremented after any save/connect dialog close to trigger sidebar refresh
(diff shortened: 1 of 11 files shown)


GOODBOY008 (approved):
@sunxiaobin89  LGTM

Issue #70, opened 2026-07-30
Feature request: automatically sync the file browser with the active terminal working directory


htazq:
## Summary

Please add an optional **Follow terminal directory** feature to the integrated
file browser.

When the user changes directories in the active SSH terminal, the file browser
below the terminal should automatically navigate to the same remote directory,
similar to MobaXterm's synchronized SFTP browser.

For example, after running:

```bash
cd /root/mlperf
```

the file browser should automatically navigate to:

```text
/root/mlperf
```

The synchronization should occur when the shell command has completed and the
prompt is displayed again. It does not need to update on every keystroke.

## Screenshot

The screenshot shows the desired synchronized state. At present, the same path
must be opened manually in the file browser after changing directories in the
terminal.

## Current behavior

The terminal and the integrated file browser maintain independent working
directories.

For example:

1. The terminal is currently in `/root/`.
2. The user runs:

   ```bash
   mkdir mlperf
   cd mlperf
   ```

3. The terminal working directory becomes:

   ```text
   /root/mlperf
   ```

4. The file browser remains in its previous directory unless the user manually
   navigates to `/root/mlperf`.

This requires users to navigate to the same directory twice and becomes
inconvenient during frequent terminal and file-transfer operations.

## Expected behavior

Add an optional synchronization mode with the following one-way relationship:

```text
Active terminal working directory -> Integrated file browser path
```

When synchronization is enabled:

1. The active terminal changes its working directory.
2. The shell command completes and the prompt returns.
3. The file browser automatically navigates to the detected directory.
4. The directory listing is refreshed automatically.

Navigating inside the file browser should **not** silently execute a `cd`
command in the terminal.

## Suggested UI

Add a toggle or link/lock button near the file-browser path bar, for example:

```text
Follow terminal directory
```

Suggested states:

- **Enabled:** the file browser follows the active terminal working directory.
- **Disabled:** the terminal and file browser can be navigated independently.
- The current state should preferably be remembered globally or per SSH
  connection.
- The UI should clearly indicate whether synchronization is active.
- Manual file-browser navigation should remain available when synchronization
  is disabled.

A suitable tooltip could be:

```text
Automatically follow the working directory of the active terminal
```

## Recommended implementation

Please avoid parsing the visible shell prompt text.

Prompt formats are not reliable because they may be:

- Customized by the user.
- Multiline.
- Colorized with ANSI escape sequences.
- Missing the current path.
- Using a shortened path.
[...]

Issue #56, opened 2026-07-22
[Bug] Edit menu commands do nothing in active terminal


htazq:
## Summary

The commands under the top **Edit** menu appear enabled when an SSH
terminal is active, but clicking them has no effect.

Affected items include:

- Copy
- Paste
- Cut
- Select All
- Find
- Find Next
- Find Previous
- Clear Screen

The keyboard shortcut for Find (`Ctrl+F`) works correctly and opens the
terminal search bar. Therefore, the terminal-side search implementation
itself is functional. The top menu commands are not connected to the
active `PtyTerminal` instance.

## Steps to reproduce

1. Start R-Shell on Windows.
2. Open an SSH connection.
3. Ensure the terminal tab is active.
4. Open **Edit**.
5. Click **Find** or another command in the Edit menu.

## Actual behavior

The menu closes, but no terminal action occurs and no error is displayed.

For example, clicking **Edit → Find** does not display the search bar,
while pressing `Ctrl+F` works correctly.

## Expected behavior

The Edit menu should operate on the currently active terminal:

- **Find** should behave the same as `Ctrl+F` and focus the search input.
- **Find Next / Find Previous** should navigate the current query.
- **Copy / Paste / Select All / Clear Screen** should invoke the
  corresponding operation on the active terminal.
- Commands that are not implemented or not applicable should be disabled
  instead of appearing clickable.

## Code analysis

`MenuBar` exposes callbacks such as `onCopy`, `onPaste`, `onSelectAll`,
and `onFind`, and binds some menu items to these callbacks.

However, `App.tsx` creates `<MenuBar>` without supplying these
terminal-operation callbacks.

Other Edit menu items, including Cut, Find Next, Find Previous, and Clear
Screen, currently have no `onClick` handler in `menu-bar.tsx`.

By contrast, `PtyTerminal` handles `Ctrl+F` locally by setting
`searchVisible`, which explains why the shortcut works while the top
menu item does not.

## Suggested fix

Add a command bridge from `App.tsx` / `MenuBar` to the currently active
`PtyTerminal`, using the existing terminal callback context or a
dedicated active-terminal command context/ref.

Only the terminal whose group and tab are currently active should execute
the command, especially when split view is enabled.

## Acceptance criteria

- Clicking **Edit → Find** behaves exactly like `Ctrl+F`.
- The search input automatically receives focus.
- Edit commands target only the active terminal in split view.
- Unsupported actions are disabled instead of silently doing nothing.
[...]

Issue #15, opened 2026-05-26
Feature request: Highlight inactive terminal tabs when new output arrives


htazq:
It would be useful if inactive terminal tabs could change color or show a distinct indicator when new output arrives. Currently the green dot mainly indicates that the SSH session is connected, but it does not clearly tell whether a background tab has produced new output. For example, when running long tasks, logs, builds, or monitoring commands in multiple SSH tabs, I need an obvious visual cue to know which inactive tab has new activity. MobaXterm handles this well by changing the tab color/indicator when there is new output, and clearing it after the user switches to that tab. Please consider adding a similar “new output/activity indicator” for background terminal tabs.

PR #71, opened 2026-07-30
feat: follow terminal cwd and download directories

### LINKED ISSUE #69: Integrated file browser cannot download remote directories
## Summary

The integrated SSH file browser supports downloading individual remote files,
but it does not provide any way to download a remote directory recursively.

This is inconsistent with the existing folder-upload capability and with the
dual-panel file transfer view, which already has recursive directory transfer
support.

## Steps to reproduce

1. Connect to a remote SSH host.
2. Open the integrated file browser.
3. Navigate to a directory containing subdirectories, for example:
   `results/20260729T055331Z`
4. Right-click a remote directory.
5. Inspect the context menu.

## Actual behavior

The directory context menu contains actions such as:

- Open Folder
- Copy / Cut
- Rename
- Duplicate
- Copy Path
- File Info
- Delete

There is no **Download Folder** or **Download Directory** action.

Selecting multiple entries does not provide a workaround because the current
multi-download implementation filters the selection to files only.

## Expected behavior

Right-clicking a remote directory should provide a **Download Folder...** action
that:

1. Lets the user select a local destination directory.
2. Recursively enumerates the remote directory tree.
3. Creates the corresponding local directory structure.
4. Downloads all files while preserving relative paths.
5. Shows transfer progress, errors, and cancellation controls.

Mixed multi-selection containing files and directories should ideally also be
supported.

## Code-level observation

In `src/components/integrated-file-browser.tsx`:

- `handleDownload()` handles only an individual file.
- `handleDownloadMultiple()` explicitly filters entries using:
  `selectedFileItems.filter(f => f.type === 'file')`
- The Download context-menu item is rendered only when:
  `file.type === 'file'`

The repository already contains recursive directory-transfer logic in the
dual-panel file browser through `FilePanel`, `DirectoryTransferDialog`, and
`onTransferDirectoryToOther`.

It would be preferable to reuse or extract that implementation so both file
browser interfaces behave consistently.

## Environment

- OS: Windows 11
- UI language: Simplified Chinese
- App version: not visible in the screenshot
- The current `main` branch also appears to lack directory download support in
  `IntegratedFileBrowser`.

htazq:
## Summary

- add an opt-out **Follow terminal directory** toggle to the integrated browser
- inject a small Bash `PROMPT_COMMAND` hook that emits standard OSC 7 working-directory reports without parsing visible prompts
- scope the latest directory to each terminal/tab and preserve the last good browser listing when a reported path is inaccessible
- add **Download directory** to the integrated browser by reusing the existing recursive transfer dialog
- let normal SSH connections open one SFTP subsystem for recursive enumeration, instead of requiring a standalone SFTP tab

## Why these issues are together

#69 and #70 meet at the integrated SSH file-browser boundary and share the same path/session isolation work. I did not bundle other open issues:

- #60 already has PR #66 and touches `pty-terminal.tsx` plus its activation tests
- PR #65 also touches `App.tsx`/locales, but fixes the unrelated connection-folder menu flow
- #56, #15, and #12 already have separate PRs

## Behavior covered

- `cd` with absolute and relative paths, `cd ..`, `cd ~`, `cd -`, `pushd`, and `popd`
- spaces, Unicode, percent encoding, `%`, `#`, and `?`
- active-tab/connection isolation and reconnect remounts
- global persisted follow toggle; file-browser navigation never sends `cd`
- inaccessible paths keep the previous path/listing and warn once until a successful load
- recursive directory creation/download, progress, per-file errors, and cancellation

## Validation

- `pnpm test` — 39 files, 525 tests passed
- `pnpm build` — passed
- `pnpm i18n:check` — 950 keys in both locales
- ESLint on every touched frontend source file — 0 errors
- `cargo test --lib --locked` in Linux/Tauri container — 108 passed, 6 ignored
- `cargo check --locked` in Linux/Tauri container — passed
- `cargo clippy --locked` — completed; only existing repository warnings
- real Docker OpenSSH test — Bash OSC 7 follow cases and same-session SFTP Unicode/nested listing passed
- `git diff --check` — passed

Repository-wide `pnpm lint` remains red on the same 3 pre-existing errors in `connection-dialog.tsx` and `settings-modal.tsx`. Repository-wide `cargo fmt --check` also remains red on pre-existing formatting outside these hunks; both CI jobs are existing baseline conditions (Rust formatting is continue-on-error).

## Known boundaries

- If `sudo -i` or `su -` starts a fresh login shell that discards `PROMPT_COMMAND`, following resumes only when that shell itself emits OSC 7 or after returning to the instrumented Bash shell. This change intentionally does not edit remote rc files, preserve unsafe environment variables through sudo, or rewrite user commands.
- Mixed file+directory multi-selection download is not included; the requested right-click directory download path is complete and reuses the existing progress/cancel UI.

Closes #69
Addresses #70

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ R-Shell combines an interactive terminal, dual-panel file manager, remote deskto
 - **Terminal search** — Regex and case-sensitive search with F3 navigation
 - **Context menu** — Copy, paste, select all, clear, save to file, reconnect
 - **IME / CJK input** — Full support for Chinese, Japanese, Korean input methods
+- **Working-directory follow** — Keep the remote file browser aligned with the active Bash terminal
 
 ### 🪟 Split Panes & Tab Groups
 
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -2904,7 +2904,50 @@ pub async fn list_local_files_recursive(
     Ok(results)
 }
 
-/// Recursively list all files/dirs under a remote directory (SFTP/FTP).
+fn walk_sftp<'a>(
+    sftp: &'a russh_sftp::client::SftpSession,
+    base: &'a str,
+    current: &'a str,
+    exclude: &'a [String],
+    results: &'a mut Vec<SyncFileEntry>,
+) -> std::pin::Pin<Box<dyn std::future::Future<Output = Result<(), String>> + Send + 'a>> {
+    Box::pin(async move {
+        let entries = crate::sftp_client::list_sftp_dir(sftp, current)
+            .await
+            .map_err(|e| e.to_string())?;
+        for entry in entries {
+            if matches_exclude(&entry.name, exclude) {
+                continue;
+            }
+            let full_path = if current == "/" {
+                format!("/{}", entry.name)
+            } else {
+                format!("{}/{}", current, entry.name)
+            };
+            let relative_path = full_path
+                .strip_prefix(base)
+                .unwrap_or(&full_path)
+                .trim_start_matches('/')
+                .to_string();
+            let is_dir = matches!(entry.file_type, FileEntryType::Directory);
+
+            results.push(SyncFileEntry {
+                relative_path,
+                name: entry.name,
+                size: entry.size,
+                modified: entry.modified,
+                file_type: entry.file_type,
+            });
+
+            if is_dir {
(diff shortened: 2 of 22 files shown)


GOODBOY008:
@sunxiaobin89 Can you help to review this PR？

PR #72, opened 2026-07-31
fix(proxy): persist proxy config so it survives saves and reconnects

sunxiaobin89:
## Summary

Proxy settings (type, host, port, username, password) configured on a connection were never persisted. Any save — creating a new connection, editing an existing one, or a failed connect that still saved the connection — silently dropped the proxy config. Reopening the Edit dialog showed empty proxy fields.

## Root cause

- The storage model (`ConnectionData`) had no proxy fields, so no save operation could store them.
- Every save/update path in `ConnectionDialog` (new connect, edit save) and the connection duplication flow omitted proxy fields.
- All `ConnectionData → ConnectionConfig` conversions in `App.tsx` (9 sites) dropped proxy fields, so the edit dialog never received them.

## Changes

- `src/lib/connection-storage.ts` — add `proxyType`, `proxyHost`, `proxyPort`, `proxyUsername`, `proxyPassword` to `ConnectionData`.
- `src/lib/connection-config.ts` (new) — `toConnectionConfig()` maps stored connections to the dialog config, carrying proxy settings; replaces 9 duplicated constructions in `App.tsx`.
- `src/components/connection-dialog.tsx` — include proxy fields in all 5 save/update calls (SSH, SFTP/FTP/Desktop, edit save); bind `value` on the proxy username input so saved data displays; normalize `proxyType` to `'none'` when loading a connection without proxy settings.
- `src/components/connection-manager.tsx` — duplication preserves proxy settings.

## Testing

- 12 new tests: `toConnectionConfig` proxy mapping, storage round-trip, and dialog persistence (edit save, failed connect keeps proxy, edit dialog shows saved values).
- Full suite: 522 tests pass, `tsc --noEmit` clean, `pnpm i18n:check` clean, no new lint issues.
- Manually verified: proxy config survives a failed connect and is shown again when editing.

> Note: proxy settings are now persisted and round-trip through the edit dialog, but are not yet forwarded to the backend during connection — that is a separate feature, not part of this fix.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,6 +13,7 @@ import { SettingsModal } from './components/settings-modal';
 import { IntegratedFileBrowser } from './components/integrated-file-browser';
 import { WelcomeScreen } from './components/welcome-screen';
 import { UpdateChecker } from './components/update-checker';
+import { toConnectionConfig } from './lib/connection-config';
 import { ActiveConnectionsManager, ConnectionStorageManager } from './lib/connection-storage';
 import { isDesktopProtocol } from './lib/protocol-config';
 import { registerRestoration, clearAllRestorations } from './lib/restoration-manager';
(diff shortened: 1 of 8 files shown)


GOODBOY008 (approved):
LGTM

PR #90, opened 2026-08-14
fix(terminal): reconnect existing SSH tab via RECONNECT_TAB

### LINKED ISSUE #89: SSH existing-tab reconnect can leave the tab stuck on "connecting"
## Summary

In `handleConnectionDialogConnect` (src/App.tsx), the **existing-tab branch** only handles SFTP/FTP and RDP/VNC. For SSH / Telnet / Raw / Serial the branch unconditionally sets the tab status to `connecting` and then does nothing further, so a tab would stay stuck on "connecting" with no terminal response.

Today no UI path reaches that branch for SSH (the dialog always issues a fresh id, editing goes through `handleSave`, the sidebar always creates a fresh session id, and quick-connect short-circuits when the tab exists), so this is a **consistency/defensive gap** rather than a currently reachable bug — but any future entry point that calls `onConnect` with an existing tab's id (e.g. a "reconnect from dialog" flow) would hit it.

## Fix

Dispatch `RECONNECT_TAB` in the existing-tab branch for SSH-family protocols, mirroring `handleReconnect`. `RECONNECT_TAB` increments `reconnectCount`, which changes `PtyTerminal`'s key in `terminal-tab-portals.tsx`, forcing a remount that opens a fresh WebSocket/PTY session on the (already established) backend SSH session. The tab status self-heals to `connected` once the new PTY reports ready.

The backend SSH session is established by the dialog's own `ssh_connect` invoke before `onConnect` fires, so no extra `ssh_connect` is needed here (unlike SFTP/FTP).

## Tests

- New regression test `src/__tests__/app-ssh-reconnect.test.tsx` (App-level, real reducer): drives the existing-tab branch with an existing tab id and asserts the terminal remounts (`RECONNECT_TAB` effect: mount count 1 → 2, `reconnectCount` 1).
- Full suite: 576 tests pass.

sunxiaobin89:
Fixes #89

## Problem

`handleConnectionDialogConnect`'s existing-tab branch only handled SFTP/FTP and RDP/VNC. SSH/Telnet/Raw/Serial tabs were set to `connecting` and never progressed because `PtyTerminal` was never remounted. Today no UI path reaches that branch for SSH (defensive gap), but any future entry point calling `onConnect` with an existing tab's id would hit the stuck state.

## Change

Dispatch `RECONNECT_TAB` for SSH-family protocols in the existing-tab branch:

```ts
} else {
  // SSH / Telnet / Raw / Serial reconnect. The backend SSH session was
  // already (re-)established by the dialog's own ssh_connect invoke, so
  // remount PtyTerminal to open a fresh WebSocket/PTY session on it.
  dispatch({ type: 'RECONNECT_TAB', tabId });
}
```

`RECONNECT_TAB` increments `reconnectCount` → `PtyTerminal` key changes in `terminal-tab-portals.tsx` → remount → fresh WebSocket/`StartPty` on the already-established backend session → status self-heals to `connected`. Mirrors the existing `handleReconnect` SSH pattern.

## Why no `ssh_connect` here

Unlike SFTP/FTP, SSH authentication already happens inside `ConnectionDialog` (`invoke('ssh_connect')` before `onConnect`), so the backend session exists by the time this branch runs.

## Tests

- New `src/__tests__/app-ssh-reconnect.test.tsx` (App-level integration, real reducer + mocked PtyTerminal): drives the existing-tab branch with the existing tab's id; asserts the terminal remounts (mount 1 → 2) and `reconnectCount` becomes 1. Verified red before the fix, green after.
- `pnpm test`: 576 tests pass; `npx tsc --noEmit` clean.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1117,6 +1117,14 @@ function AppContent() {
             description: error instanceof Error ? error.message : String(error),
           });
         }
+      } else {
+        // SSH / Telnet / Raw / Serial reconnect. The backend SSH session was
+        // already (re-)established by the dialog's own ssh_connect invoke, so
+        // remount PtyTerminal to open a fresh WebSocket/PTY session on it.
+        // (RECONNECT_TAB increments reconnectCount, changing PtyTerminal's key
+        // in terminal-tab-portals so it remounts; the tab status self-heals to
+        // 'connected' once the new PTY session reports ready.)
+        dispatch({ type: 'RECONNECT_TAB', tabId });
       }
     } else {
       if (isDesktop) {
--- a/src/__tests__/app-ssh-reconnect.test.tsx
+++ b/src/__tests__/app-ssh-reconnect.test.tsx
@@ -1,0 +1,308 @@
+/**
+ * Regression test for: SSH existing-tab reconnect stuck on "connecting".
+ *
+ * Root cause (App.tsx handleConnectionDialogConnect): the existing-tab branch
+ * only handled SFTP/FTP and RDP/VNC. SSH/Telnet/Raw/Serial tabs were set to
+ * 'connecting' and never progressed — the terminal was never remounted, so no
+ * fresh WebSocket/PTY session was started.
+ *
+ * Fix: dispatch RECONNECT_TAB in the existing-tab branch for SSH-family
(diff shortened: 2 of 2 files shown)


GOODBOY008 (approved):
LGTM

PR #101, opened 2026-08-20
feat(i18n): localize remaining hardcoded user-facing strings

sunxiaobin89:
## Summary

Completes the project-wide i18n pass started in #83 by routing the last hardcoded user-facing strings through `react-i18next`, and fixes the remaining English items in the macOS native menu bar.

**Commit 1 — hardcoded strings** (`e96329f`): a full audit of every page (menus, context menus, dialogs, terminal, welcome screen, restore overlay) found the remaining untranslated strings and localizes them all.

**Commit 2 — native macOS menu** (`aebb3ec`): the app-menu predefined items (About / Services / Hide / Hide Others / Show All / Quit) were created without text and fell back to muda's hardcoded English strings; they now follow the in-app language like every other item. The bundle also declares `CFBundleLocalizations` (en, zh-Hans, zh-Hant) + a `zh-Hans.lproj` so macOS treats r-shell as supporting Chinese.

## macOS native menu — what follows which language

| Menu items | Language source |
|---|---|
| All app-created items (File/Edit custom items, Tools, Connection, Window, and the predefined Undo/Redo/Cut/Copy/Paste/Select All with explicit text) | Follow the **in-app** language, switching live |
| macOS system-injected Edit items (AutoFill, Start Dictation, Emoji & Symbols) | Follow the **system** language — they are injected by macOS and cannot follow the in-app language. This is accepted macOS behavior (see [Tutanota issue #6221](https://github.com/tutao/tutanota/issues/6221), marked wontfix "intended behaviour"; VS Code behaves the same). Documented in a comment at the Edit menu. |
| Services submenu contents | Provided by other apps' services; follow the system language |

## Windows / Linux

Unaffected and already correct: no native menu (the web MenuBar is fully `react-i18next` driven), so it always follows the in-app language (Auto-follow-system / English / 简体中文). The Info.plist / `.lproj` / native-menu changes are macOS-only (`#[cfg(target_os = "macos")]`).

## Verification

- `pnpm i18n:check` — key parity (1081 keys in both files)
- `pnpm lint` — no new warnings vs baseline (total warnings dropped 240 → 236)
- `pnpm test` — 583/583 passed (includes the "every `t()` key exists in both locales" guard)
- `pnpm build` + `pnpm tauri build` — pass; fresh bundle contains `CFBundleLocalizations` + `zh-Hans.lproj` and embeds the new keys
- Manual test on macOS: Chinese mode fully Chinese; English mode switches all app-controlled items to English (system-injected items stay in the system language, per the decision above)

## Not translated (per project convention)

Protocol names, font/theme names, keyboard symbols, units, backend/Rust error messages, and dead/unmounted components (`sftp-panel.tsx`, legacy `terminal.tsx`).


src/locales/en.json
```diff
@@ -394,6 +402,7 @@
     "newConnection": "New Connection",
     "newConnectionDesc": "Connect via SSH, SFTP, FTP, or FTPS",
     "connectionManager": "Connection Manager",
+    "connectionManagerHighlight": "Left sidebar ⌘B",
```
GOODBOY008: **nit (i18n): don't bake the keybinding into the localized string**

`"Left sidebar ⌘B"` mixes two different things that should stay separate:

1. **showname** — the label (`"Left sidebar"`) — this is what belongs in the locale file and gets translated (`.connectionManagerHighlight` → `"左侧栏"`).
2. **shortcut** — the keybinding (`Ctrl+B`) — OS-specific, should never live in a translation file, and must be converted per-platform (⌘ on macOS, Ctrl/Ctrl+B on Windows/Linux).

This repo already has the exact helper + a live demo — please follow it:
- `src/lib/keyboard-shortcuts.ts:85` — `formatKeyboardShortcut(shortcut, isMac)` turns `'Ctrl+B'` into `⌘B` on macOS and `Ctrl+B` elsewhere.
- `src/components/menu-bar.tsx:126-127` — `const isMac = navigator.platform.toUpperCase().indexOf('MAC') >= 0;` + `formatShortcut = (s) => formatKeyboardShortcut(s, isMac);`
- Use `DEFAULT_LAYOUT_SHORTCUTS.toggleLeftSidebar` (`'Ctrl+B'`) from the same module as the source.

Suggested fix: in `welcome-screen.tsx`, keep `t('welcome.connectionManagerHighlight')` for the name only and render the shortcut separately, e.g. `formatKeyboardShortcut(DEFAULT_LAYOUT_SHORTCUTS.toggleLeftSidebar, isMac)` inside the existing `{action.highlight && ...}` element (or as a `shortcut` badge consistent with the other two cards), then drop `⌘B` from both locale files.

While you're in there: `welcome-screen.tsx` lines 41/57 hardcode `⌘N` / `⌘,` — same OS-conversion problem; switching those to the helper would make Windows/Linux show `Ctrl+N` / `Ctrl+,` too.

GOODBOY008 (changes requested):

src/locales/en.json
sunxiaobin89: Fixed in 678997e. The shortcut is no longer baked into the localized string: `welcome.connectionManagerHighlight` now holds only the label ("Left sidebar" / "左侧栏"), and the ⌘B badge is rendered separately via `formatKeyboardShortcut(DEFAULT_LAYOUT_SHORTCUTS.toggleLeftSidebar, isMac)` so Windows/Linux show Ctrl+B. Also routed the New Connection (⌘N → Ctrl+N) and Preferences (⌘, → Ctrl+,) hints through the same helper as suggested. Verified: i18n parity (1081 keys), tsc, lint, 583 tests.

GOODBOY008 (approved):

Issue #95, opened 2026-08-18
[Bug] SSH terminal disconnects automatically and WebSocket reconnect cannot recover the PTY session


htazq:
## Summary

R-Shell SSH terminal sessions sometimes disconnect automatically during normal use, even when the remote server itself is still reachable.

The most obvious symptom is:

```text
[Error: Connection lost: PTY connection closed]

[Connection lost. Reconnecting in 2s (attempt 1/5)...]

✓ WebSocket connected

[Error: Error handling message: Channel send error]

[Error: Error handling message: PTY connection not found]

[Connection failed permanently. Use right-click → Reconnect to retry.]
```

The important part is that the **WebSocket reconnects successfully**, but the terminal still cannot recover because the original PTY session is already gone.

This makes R-Shell noticeably less reliable than clients such as MobaXterm for long-running SSH sessions.

## Actual scenario

Environment:

```text
Client OS: Windows
Remote OS: Ubuntu Linux
Protocol: SSH
Terminal mode: WebSocket + PTY
Remote host example: 172.18.35.109
```

I was running:

```bash
fstrim -v /data/storage1
```

The first execution completed normally:

```text
/data/storage1: 11.8 TiB (12995154272256 bytes) trimmed
```

I then ran the same command again.

During the command, without manually disconnecting the SSH session, R-Shell suddenly reported:

```text
[Error: Connection lost: PTY connection closed]
```

R-Shell automatically attempted to reconnect:

```text
[Connection lost. Reconnecting in 2s (attempt 1/5)...]

✓ WebSocket connected
```

However, immediately after the WebSocket was restored:

```text
[Error: Error handling message: Channel send error]

[Error: Error handling message: PTY connection not found]
```

and finally:

```text
[Connection failed permanently. Use right-click → Reconnect to retry.]
```

So the reconnect mechanism does not actually restore the existing terminal session.

## Screenshot

The attached screenshot captures the entire failure sequence:

```text
fstrim
    ↓
PTY connection closed
    ↓
automatic reconnect
    ↓
WebSocket connected
    ↓
Channel send error
    ↓
PTY connection not found
    ↓
permanent failure
```

## Why this looks like an R-Shell PTY / WebSocket lifecycle problem

The failure does not look like a simple SSH authentication or TCP connection failure.

R-Shell explicitly reports:

```text
✓ WebSocket connected
```

after the disconnect.

But after the WebSocket reconnects, the backend responds with:

```text
PTY connection not found
```

[...]

PR #104, opened 2026-08-23
fix(terminal): recover SSH-dead sessions via typed errors and auto re-auth

GOODBOY008:
## Problem (from #95)

The reported failure sequence was:

```
[Error: Connection lost: PTY connection closed]   ← SSH channel/transport died first
[Connection lost. Reconnecting in 2s (attempt 1/5)...]
✓ WebSocket connected                              ← transport rebuilt
[Error: Error handling message: Channel send error] ← StartPty on the STALE SshClient (russh SendError) — ignored by the keyword filter
[Error: Error handling message: PTY connection not found] ← symptom misclassified as the verdict
[Connection failed permanently.]
```

Root cause: the auto-reconnect assumes the SSH session outlives the WebSocket. When the failure order is reversed (SSH dies first), the stale `SshClient` is never evicted (`is_connected` has no callers), `StartPty` fails on the corpse, the fatal `Channel send error` matches none of the frontend's keyword filters, and the follow-on `PTY connection not found` hard-caps retries — even though a full re-auth (right-click → Reconnect) works fine.

## Changes

**Backend**
- `start_pty_connection` returns `PtyStartError::SshSessionDead` when the SSH session is missing or the transport is dead (`SendError` / `Disconnect` / `HUP` / timeout class — verified against russh 0.44.1). Transient `ChannelOpenFailure` (e.g. MaxSessions) does **not** evict.
- Dead sessions are evicted with an `Arc`-identity guard (a concurrently recreated connection survives), including PTY teardown and OS-info cache cleanup.
- `WsMessage::Error` carries an optional machine-readable `code` (`ssh_session_dead`); absent on legacy errors — fully backward compatible on the wire.

**Frontend**
- `PtyTerminal` reacts to the coded error by stopping both WS retry loops and escalating once to the App-level full reconnect (`onReconnectTab`: disconnect + re-authenticate + remount). A module-level budget (2 per connection per 10 min, survives remounts) prevents pathological loops; beyond it, the manual-reconnect hint is shown.
- Legacy uncoded errors keep the existing keyword behaviour unchanged.
- New i18n keys in en + zh-CN (`sshSessionLost`, `sshSessionDeadPermanent`).

## Tests
- Rust (6 new): error classification (dead vs MaxSessions-transient vs unrelated), missing-connection → `SshSessionDead`, eviction identity guard, serde round-trip with/without `code`, `WsError` code mapping. `cargo test`: 141 passed.
- Frontend (4 new): escalation fires once with status/close/message assertions, no double-escalation per mount, legacy path unchanged, budget exhaustion across remounts. Full suite: 593 passed. `tsc --noEmit` clean, i18n parity 1082/1082, eslint 0 errors on touched files.

## Scope

Partially addresses #95 — the failure order observed in the report (SSH dies first) is now recoverable end-to-end. The complementary scenario (pure WebSocket drop with SSH alive → PTY grace period + reattach) is intentionally left as a follow-up on top of the detached-session registry from #76.

### DIFF
--- a/src-tauri/src/connection_manager.rs
+++ b/src-tauri/src/connection_manager.rs
@@ -12,6 +12,51 @@ use tokio::sync::mpsc;
 use tokio::sync::RwLock;
 use tokio_util::sync::CancellationToken;
 
+/// Error from starting a PTY session, distinguishing a dead/unusable SSH
+/// session (the frontend must re-authenticate — a WebSocket retry cannot
+/// recover) from other failures that leave the session intact.
+#[derive(Debug)]
+pub enum PtyStartError {
+    /// The SSH session is gone or unusable. Any stale client has already
+    /// been evicted from the connection map.
+    SshSessionDead(String),
+    /// Any other failure; the session is left as-is.
+    Other(anyhow::Error),
+}
+
+impl std::fmt::Display for PtyStartError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            PtyStartError::SshSessionDead(msg) => write!(f, "{}", msg),
+            PtyStartError::Other(e) => write!(f, "{}", e),
+        }
+    }
+}
+
+impl std::error::Error for PtyStartError {}
+
+impl From<anyhow::Error> for PtyStartError {
+    fn from(e: anyhow::Error) -> Self {
+        PtyStartError::Other(e)
+    }
+}
+
+/// Whether an error raised while opening a channel indicates the underlying
+/// SSH session itself is dead (transport gone), as opposed to recoverable
+/// conditions like the server temporarily refusing more channels
+/// (`ChannelOpenFailure`, e.g. MaxSessions exhaustion).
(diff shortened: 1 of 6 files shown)


GOODBOY008:
Thanks @copilot-pull-request-reviewer — went through all three comments in 6528698:

1. **`ws.close()` → `onclose` interference** — valid, addressed. The escalation set both retry counters to MAX, so no retries were actually scheduled, but `onclose` *did* print an `autoReconnectFailed` / `reconnectFailedPermanently` banner directly contradicting the escalation message. `onclose` now early-returns once `sshDeadEscalatedRef` is set (App.tsx owns recovery at that point).
2. **Unbounded escalation map** — valid, addressed. Entries now prune opportunistically (only when the map exceeds 64 entries, and only expired windows), since `-dup-<ts>` ids make stale entries accumulate.
3. **Mock never exercises `onclose`** — valid, addressed. The mount helper now follows the real `Success` → `PtyStarted` flow (so `hasEverConnected` is set), and the escalation test explicitly fires `onclose` after the coded error, asserting no auto-reconnect banners and no extra App-level reconnect — exactly the regression class you flagged.

Issue #87, opened 2026-08-13
[Bug] Multiple long-lived SSH tabs can become blank and input-dead after being left overnight


htazq:
## Summary

On Windows, after leaving R-Shell open overnight with many SSH tabs, multiple terminal panes can become completely blank/black and stop showing any response to keyboard input.

This does not appear to be a complete SSH or application failure. In the observed case, the active terminal tab was still shown as **Connected**, while the integrated SFTP browser remained usable and the System Monitor continued updating live data. The R-Shell UI itself remained responsive.

Switching between affected SSH tabs did not recover the terminal, and pressing `Ctrl+L` produced no visible response.

The problem was observed with roughly 10+ SSH tabs open across multiple Linux hosts.

## Observed behavior

Two different active SSH sessions reproduced the same condition.

One affected tab was connected to `172.18.35.108`; another affected tab was connected to `172.18.35.103`.

In both cases the main terminal viewport was blank, while the SFTP panel displayed remote directory contents and the System Monitor continued to update.

This makes a single remote-host SSH failure unlikely and points to the terminal renderer, tab activation path, or PTY output path.

## Related existing issues

This is related to #7 / #13, which addressed terminals becoming stale or unfocused after tab switching, but the current activation path may still contain a race.

Inactive terminal portal hosts remain mounted and use:

```tsx
style={{ display: isActive ? 'block' : 'none' }}
```

When a tab becomes active, `PtyTerminal` performs one `requestAnimationFrame()` and then:

```ts
if (container.offsetWidth <= 0 || container.offsetHeight <= 0) return;

fitAddon.fit();
term.refresh(0, term.rows - 1);
term.focus();
```

If the container is still zero-sized during that single frame, activation returns without retrying the refresh/focus operation.

The current activation tests force `offsetWidth = 800` and `offsetHeight = 600`, so a delayed `0x0 -> visible` transition is not covered.

The problem is also related to #60 and PR #66.

Current main still resets and clears xterm after 2 MiB of cumulative decoded output:

```ts
term.reset();
term.clear();
```

PR #66 also demonstrates that current PTY output-credit accounting can grow the backend permit window over time during low-rate sustained output.

These issues are highly relevant to overnight workloads such as `top`, `watch`, `iostat`, `tail -f`, benchmark logs and similar commands, but they do not by themselves fully explain a connected terminal becoming completely blank and apparently input-dead.

## Possible WebGL involvement

[...]

PR #105, opened 2026-08-23
fix(terminal): lazy WebGL per visible pane + deterministic activation repair

GOODBOY008:
## Root cause (from the #87 symptom set)

Blank + input-dead panes with the badge still **Connected**, SFTP working, System Monitor updating, and tab switching never recovering localizes the failure to the xterm render layer (the WS is loopback and can't silently die; the status badge is React state driven only by WS events). Two defects combined:

1. **One WebGL context per mounted terminal, held for life — including hidden panes.** 10+ tabs exceeded Chromium's live-context budget; the oldest contexts got evicted overnight. Under WebGL, text exists *only as pixels on the GL canvas*, so an evicted pane renders black forever while keystrokes still flow into the dead context — which is exactly why Ctrl+L showed *no visible response* while everything else stayed alive. (Same failure class as xtermjs/xterm.js#2253 / #6015.)
2. **One-shot activation.** The latch was consumed *before* a single rAF that bailed permanently on a 0×0 container, and the ResizeObserver path only ever called `fit()` — never `refresh()` — so tab switching repainted *through the still-registered dead renderer*: black on black.

## Changes (`pty-terminal.tsx`, frontend-only)

- **Lazy WebGL**: the addon loads when a pane becomes visible and is disposed when it hides. Live contexts drop from *mounted panes* to *visible panes* (1–4), below eviction pressure. Hidden panes keep their text via the DOM renderer.
- **Hardened context loss**: dispose wrapped in try/catch, then forced `fit()` + `refresh(0, rows-1)` so the fallback renderer repaints immediately instead of relying on the addon's best-effort dispose path.
- **Deterministic activation**: retries every animation frame until the container has a real size (~2s budget), and only then consumes the latch, ensures a working renderer, and runs fit + full refresh + focus. `wasActiveRef` now starts `false`, so **mount-active panes pass through the same measured path** — previously a terminal that mounted active behind a 0×0 container had no retry at all (a hole the new regression test exposed during development).
- The ResizeObserver completes a still-pending activation on a 0×0 → non-zero transition (wake-from-display-sleep coverage).

## Tests (5 new, `pty-terminal-renderer-lifecycle.test.tsx`)

- Inactive mount creates **no** WebGL context
- Activation loads WebGL + repaints + focuses; deactivation disposes it
- Active mount loads WebGL
- Context-loss handler disposes and force-repaints via fit + full refresh
- **0×0 → sized retry**: five frames at 0×0 (no repaint, latch not consumed) → container becomes visible → fit + `refresh(0, rows-1)` + focus land — the exact regression #87 asked for

Full suite: 54 files / 594 tests pass. `tsc --noEmit` clean, eslint 0 errors on touched files.

## Notes

- Addresses the renderer root cause of #87; the sustained-output bounding (#60 half) already landed in #66, and the per-subsystem health-state tracking from the issue remains open follow-up work.
- Stacked on current `main` (93d37c9). Overlaps trivially with #104's `pty-terminal.tsx` edits (different regions); will rebase if #104 merges first.

### DIFF
--- a/src/__tests__/pty-terminal-renderer-lifecycle.test.tsx
+++ b/src/__tests__/pty-terminal-renderer-lifecycle.test.tsx
@@ -1,0 +1,379 @@
+import React from 'react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { act, cleanup, render } from '@testing-library/react';
+import { PtyTerminal } from '../components/pty-terminal';
+
+const mocks = vi.hoisted(() => {
+  const terminals: Array<any> = [];
+  const webSockets: Array<any> = [];
+  const webglInstances: Array<any> = [];
+  const terminalCallbacks = {
+    onReconnectTab: vi.fn(),
+    onWorkingDirectoryChange: vi.fn(),
+  };
+
+  class MockTerminal {
+    cols = 80;
+    rows = 24;
+    options: Record<string, unknown> = {};
+    buffer = {
+      active: {
+        length: 0,
+        getLine: vi.fn(),
+      },
+    };
+    parser = {
+      registerOscHandler: vi.fn(() => ({ dispose: vi.fn() })),
+    };
+
+    loadAddon = vi.fn();
+    open = vi.fn();
+    focus = vi.fn();
+    refresh = vi.fn();
+    writeln = vi.fn();
+    write = vi.fn((_data: string, callback?: () => void) => callback?.());
+    onSelectionChange = vi.fn(() => ({ dispose: vi.fn() }));
+    onLineFeed = vi.fn(() => ({ dispose: vi.fn() }));
+    attachCustomKeyEventHandler = vi.fn();
+    onData = vi.fn(() => ({ dispose: vi.fn() }));
+    onResize = vi.fn(() => ({ dispose: vi.fn() }));
+    hasSelection = vi.fn(() => false);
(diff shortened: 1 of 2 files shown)


GOODBOY008:
Thanks @copilot-pull-request-reviewer — one comment adopted, one corrected with evidence (423290b):

**Adopted — `>100px` gate on activation completion.** Correct catch: the ResizeObserver backstop inherited `debouncedFit`'s threshold, so a pane that exhausted its rAF budget at 0×0 and then appeared at a sub-100px size (deep splits are legitimately that narrow) would never activate — the same permanently-blank class this PR fixes. Activation now triggers on **any non-zero** size while `debouncedFit` keeps its 100px threshold, and a new regression test covers the exact scenario (budget exhausted → 30px transition → fit + full refresh + focus must land).

**Corrected — "fallback is not DOM-based".** Verified against the installed `@xterm/xterm` 6.0.0: core's `_createRenderer()` constructs `c.DomRenderer` — the DOM renderer *is* built into v6 core and *is* what the WebglAddon restores on dispose (its dispose handler calls `setRenderer(terminal._core._createRenderer())` + `handleResize`, which triggers a full refresh). So hidden panes do keep their text as DOM rows; that's the load-bearing property this PR relies on. I've sharpened both comments to cite that mechanism so the next reader doesn't have to re-derive it.

PR #108, opened 2026-08-23
feat(terminal): per-subsystem session health tracking

GOODBOY008:
Final item from #87's acceptance criteria: track the terminal pipeline's subsystems separately and reconcile tab status against them, so a **Connected** badge can't outlive a dead terminal pipeline (the overnight scenario where SFTP and the system monitor kept the shared SSH session alive while the PTY was gone).

## Backend
- `SessionHealth` + `ConnectionManager::session_health()` — point-in-time snapshot per connection: `sshConnected`, `hasPty`, `ptyGeneration`, `detached`, `connectionType`. Locks taken sequentially, unknown ids report all-false.
- New `get_session_health` Tauri command.

## Frontend
- `src/lib/session-health.ts` — the reconciler as a testable helper: for every connected SSH terminal tab, fetch health and downgrade when the SSH session is gone **or** the PTY is missing with nothing parked. A parked PTY (Ctrl+A+D or a WS drop inside the grace window) is healthy — alive, just not streaming. Fetch failures are skipped so a momentarily unreachable backend can't flap tabs.
- `App.tsx` polls every 30s against the latest tab snapshot (ref-held, so the interval survives group-state churn).

## Tests
- Rust: map-presence semantics (unknown → all false; per-subsystem population). `cargo test`: 151 passed.
- Frontend: reconciler unit tests — SSH-dead, PTY-dead-no-park, parked-is-healthy, tab filtering, fetch-failure tolerance. Full suite: 56 files / 608 tests; `tsc` clean.

Built on top of #76 (detach registry) / #106 / #107 — all merged.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -191,6 +191,17 @@ pub async fn ssh_disconnect(
     }
 }
 
+/// Per-subsystem health snapshot for a connection's terminal pipeline
+/// (SSH alive / PTY attached / generation / detached). Used by the frontend
+/// to reconcile tab status so "Connected" can't outlive a dead PTY.
+#[tauri::command]
+pub async fn get_session_health(
+    connection_id: String,
+    state: State<'_, Arc<ConnectionManager>>,
+) -> Result<crate::connection_manager::SessionHealth, String> {
+    Ok(state.session_health(&connection_id).await)
+}
+
 /// List connection IDs that currently have a detached (background) session.
 #[tauri::command]
 pub async fn list_detached_sessions(
--- a/src-tauri/src/connection_manager.rs
+++ b/src-tauri/src/connection_manager.rs
@@ -6,6 +6,7 @@ use crate::sftp_client::StandaloneSftpClient;
 use crate::ssh::{PtySession, SshClient, SshConfig};
 use crate::vnc_client::VncClient;
 use anyhow::Result;
+use serde::Serialize;
 use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Duration;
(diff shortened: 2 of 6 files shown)


GOODBOY008:
Thanks @copilot-pull-request-reviewer — went through all three comments in fb85c24. Two adopted, one corrected with evidence:

**Adopted — overlapping polls.** Correct: an in-flight pass could overlap the next 30s tick. `reconcileSessionHealth` now skips (returns []) while a pass is running, released in a `finally` so a failure can't wedge the guard. Regression test covers it (second pass skipped, fetch invoked exactly once, first pass still completes).

**Adopted — protocol filtering.** Tightened `isSshTerminalTab` from "anything that isn't SFTP/FTP" to explicit SSH-or-legacy-undefined, and added rejection tests. Note on the premise: this codebase's protocols are SSH/SFTP/FTP/RDP/VNC (Telnet/Raw/Serial don't exist here) and RDP/VNC tabs are `tabType: 'desktop'`, so they were already excluded by the tabType check — the explicit protocol gate is defense-in-depth rather than fixing an active misclassification.

**Corrected — IPC parameter-name mismatch.** Verified against tauri-macros 2.6.2: the default `ArgumentCase::Camel` (`wrapper.rs` line 51) lower-camelCases Rust parameters, and the runtime does an exact-key lookup (`command.rs`: `v.get(self.key)` → "missing required key"). `{ connectionId }` therefore binds to `connection_id: String` — and this is empirically proven by the shipped System Monitor, which invokes `get_system_stats(connection_id)` with `{ connectionId }` and works. (The snake_case invoke in `terminal.tsx` is the real anomaly, pre-existing and out of scope here — worth its own issue.)

PR #79, opened 2026-08-08
feat(ssh-tunnel): support SSH tunnel (jump host) when creating a connection

GOODBOY008:
## Summary

Adds **SSH tunnel (jump host / bastion)** support to connection creation, the "basic feature" requested in **Closes #41**. Connections can now be routed through an intermediate SSH server — for the terminal (SSH) and the file browser (SFTP).

## What changed

**Backend (Rust)**
- `ssh/mod.rs`: new `TunnelConfig` (`host`/`port`/`username`/`auth_method`) on `SshConfig`; `connect_via_ssh_tunnel` opens a direct-tcpip channel through the jump host and hands it to russh's `connect_stream` (tunnel → proxy → direct precedence). Auth logic extracted into a shared `authenticate_session` helper used by both target and jump host.
- `commands.rs`: `tunnel_*` fields on `ConnectRequest` / `SftpConnectRequest` + shared `build_tunnel_config`.
- `sftp_client.rs`: `SftpConfig.tunnel` + tunnel routing so the **file browser connects through the same jump host**.

**Frontend**
- New **SSH Tunnel** tab in the connection dialog: enable toggle + host / port / username + password **or** private-key auth (reuses `PasswordInput`).
- Tunnel fields persisted on `ConnectionConfig` / `ConnectionData` (edit round-trip, failed-connect survival).
- `buildSshConnectRequest` and new `buildSftpConnectRequest` carry the tunnel fields; all 7 inline `sftp_connect` request objects in `App.tsx` centralized through the helper (−71 lines).
- i18n keys added in both `en.json` and `zh-CN.json`.

## Tests

- Rust: tunnel config + `build_tunnel` unit tests; `cargo test` 141 passing.
- Frontend: request-builder tunnel cases, dialog tunnel persistence, storage round-trip; `pnpm test` 576 passing.
- `tsc`, eslint (no new errors), i18n parity all clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #114, opened 2026-08-26
feat(security): encrypt stored credentials at rest with keychain-backed master key

### LINKED ISSUE #100: [Security] SSH passwords stored in plaintext in local WebView localStorage
## Problem

SSH connection passwords, key passphrases, and proxy/VNC credentials were stored in plaintext in the app's WebView `localStorage`. On a desktop app this is a lower-risk vector than a browser, but the credentials were still readable on disk, and the export feature serialized them into files that could be shared.

## Solution

Credentials are now encrypted at rest with an app-level master key.

- A random 32-byte **master key** is generated on first use and kept in the **OS keychain** (Rust `keyring` crate). There is one keychain entry for the whole app, so the user is prompted once rather than once per credential.
- Each secret field is encrypted with **AES-256-GCM** and persisted as `v1:<base64 nonce>:<base64 ciphertext>`. Plaintext secrets exist only in memory transiently during a connection and are never written to `localStorage`.
- All connection writes flow through a single persistence path that drops any non-encrypted secret, as a defensive safeguard.
- On launch, previously-stored plaintext secrets are migrated to the encrypted format. If a migration fails, the plaintext copy is left in place so the only copy is never lost.
- When editing a connection, a stored secret is never echoed back into the form: the field stays empty with a hint to retain the saved value. Leaving it blank keeps the stored secret; typing a value replaces it.
- Export omits all secret fields from exported bundles, so exported config files never contain credentials (encrypted or otherwise).

Encrypted fields: `password`, `passphrase`, `proxyPassword`, `vncPassword`.

## Design note

Per-credential keychain access and export-only sanitization were considered and not adopted: the former would have raised an authorization prompt for every stored credential, and the latter would have left credentials readable on disk at rest. The chosen design aligns with the mainstream approach used by SSH clients — at-rest encryption with a keychain-backed master key.

sunxiaobin89:
## Summary

Encrypts connection secrets (SSH password, key passphrase, proxy password, VNC password) at rest. Previously these were stored in plaintext in the app's WebView `localStorage`; the export feature also wrote them into shareable files.

## Approach

Single app-level master key + AES-256-GCM:

- A random 32-byte **master key** is generated on first use and stored in the **OS keychain** (Rust `keyring` crate, one entry for the whole app → one authorization prompt, never per-credential).
- Secrets are persisted as `v1:<b64 nonce>:<b64 ciphertext>`. Plaintext only exists in memory transiently at connect time — **never written to `localStorage`**.
- **`persistConnections()`** is the single write path and strips any non-`v1:` plaintext secret as a defensive layer.
- Startup migration seals legacy plaintext already in storage (keeps the plaintext copy on failure so the only copy is never destroyed).
- Edit dialog never echoes a stored secret: the field stays empty with a hint ("leave blank to keep the saved password"); blank = keep stored, typed = replace.
- **Export strips all secret fields** (plaintext and ciphertext) from shared config bundles.

## Verification

- 636 frontend tests + 152 Rust tests pass.
- `tsc --noEmit` clean; i18n key parity (en/zh-CN) holds.
- Manual test of save / edit / connect / reconnect / migration round-tripped end to end.

Fixes #100

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -133,6 +133,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
+name = "apple-native-keyring-store"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2b350bfd03649e07aa05c0a81b3e15934374e585c98204a57e20b9d49f49bb9a"
+dependencies = [
+ "keyring-core",
+ "log",
+ "security-framework",
+]
+
+[[package]]
 name = "arbitrary"
 version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2722,6 +2733,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "keyring"
+version = "4.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72585bb6cc9bc370d1d545b7e23fcce71dfd4461c5e15275e3cf51bdfd9a980a"
+dependencies = [
+ "apple-native-keyring-store",
+ "keyring-core",
+ "windows-native-keyring-store",
+ "zbus-secret-service-keyring-store",
+]
+
+[[package]]
+name = "keyring-core"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fb1e621458ca9c51aa110bd0339d4751a056b9576bf1253aee1aa560dda0fc9d"
+dependencies = [
+ "log",
+]
+
+[[package]]
 name = "kv-log-macro"
 version = "1.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3024,6 +3056,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "num"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "35bd024e8b2ff75562e5f34e7f4905839deb4b22955ef5e73d2fea1b9813cb23"
+dependencies = [
+ "num-bigint",
+ "num-complex",
+ "num-integer",
+ "num-iter",
+ "num-rational",
+ "num-traits",
+]
+
+[[package]]
 name = "num-bigint"
 version = "0.4.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3051,6 +3097,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "num-complex"
+version = "0.4.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "73f88a1307638156682bada9d7604135552957b7818057dcef22705b4d509495"
+dependencies = [
+ "num-traits",
+]
+
+[[package]]
 name = "num-conv"
 version = "0.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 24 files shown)


sunxiaobin89:
@GOODBOY008 Friendly ping on this one — happy to adjust or rebase as needed. 🙂

One thing I noticed: since this PR was opened, `main` gained SSH tunnel / jump-host support (#79), which introduces a new credential field (tunnel password). If the overall approach here looks acceptable, I'd suggest including that field in the encryption scope as well — I'm happy to fold it in together with the rebase onto the latest `main`.

sunxiaobin89:
Rebased onto the latest `main` (v2.9.0) — the conflicts with the new quit-guard commands and the SSH tunnel tab are resolved.

While rebasing, I also folded the new SSH tunnel (jump host) fields from #79 into the encryption scope: `tunnelPassword` and `tunnelPassphrase` are now sealed like the other secrets, never echoed back in the edit dialog, and stripped from exported configs. Any legacy plaintext tunnel credentials migrate on startup together with the other secret fields.

Verification: 718 frontend tests + 170 Rust tests pass; `tsc --noEmit`, `cargo clippy` and i18n key parity are clean.

sunxiaobin89:
Rebased again onto `main` to pick up #125 (passwordless hosts) and #103 (default SSH key path). The credential checks now follow the new passwordless semantics — blank passwords and empty key paths stay valid, while the stored-secret "keep on blank" behavior of this PR is unchanged. `openConnectionSecrets` runs before the shared `connectionHasCredentials` check so restore-time decisions see decrypted values.

Verification: 732 frontend tests + 177 Rust tests pass; `tsc --noEmit` and i18n key parity are clean.

sunxiaobin89:
Thanks for the thorough review — verified each finding against the code and fixed them all. Rebased onto latest `main` (#127) in the same pass. Summary:

**Fixed (all confirmed real):**

1. **Legacy migration destroyed plaintext** (connection-storage.ts:152): `migrateFromSessionStorage` no longer routes through `persistConnections`; it writes migrated records verbatim so the startup seal migration can encrypt them. +tests
2. **Seal-failure credential loss** (App.tsx:330 / connection-storage.ts:113): added `markSealFailed`/`clearSealFailed` — persistence skips stripping for connections whose seal attempt failed, so no unrelated write (`updateLastConnected`, folder ops) can destroy the only plaintext copy; protection lifts once sealing succeeds. +tests
3. **previousSecrets leak across connections** (connection-dialog.tsx:229): now reset when the dialog opens for a new connection and when it closes. +tests
4. **Connect used blank secrets** (connection-dialog.tsx:389): `handleConnect` now builds a connect-only config that decrypts retained sealed secrets for blank fields (plaintext transient in memory only; persistence keeps the sealed form). Note: edit mode's Save→`onSave`→`App.handleSaveConnection` path already decrypted retained secrets — this closes the dialog-side `handleConnect` path symmetrically. +tests
5. **Nonce length panic** (commands.rs:3652): `credential_open` now returns a proper error when the decoded nonce is not 12 bytes.
6. **Trimmed secrets** (connection-dialog.tsx:337): secrets are sealed as typed; whitespace is preserved (the connect request already sent the untrimmed value — storage now matches).
7. **Sealed-format validation parity** (connection-storage.ts:112): the persistence filter now uses `isSealed()` (3-segment check) instead of a bare `startsWith('v1:')` prefix. +tests
8. **Profile storage** (config-export-import.ts:110 / App.tsx:318): startup migration now seals legacy plaintext profile passwords as well, and `importProfiles` strips secret fields from imported bundles (imports never persist outside secrets; storage secrets go through the migration). +tests

**On severity for #8:** `saveProfile`/`updateProfile` are only reachable from the unused `_handle*` dead code in connection-dialog.tsx, and current exports already strip all `SECRET_FIELDS` — so the live risk was limited to legacy stores / pre-sanitization bundles. Both paths are now covered regardless.

Verification: 753 frontend tests (10 new) + 177 Rust tests pass; `tsc --noEmit` and i18n parity clean; manual test of save / edit / connect / migration round-trip on the DMG build.

GOODBOY008 (approved):

PR #119, opened 2026-08-28
fix(file-viewer): focus-aware Ctrl+W, single-window reuse and persistent editor windows

GOODBOY008:
## What

Fixes Ctrl+W closing the wrong thing and hardens the file-editor window's state maintenance.

### Bug: Ctrl+W while editing closes the active terminal tab

The main window registers Ctrl+W → *close active terminal tab* as an OS-level global shortcut. When the file-viewer window (a separate Tauri webview window) has focus, the main window's blur handling kept the shortcut registered (intended so shortcuts fire while the app is backgrounded) — but it could not tell "another app focused" from "a sibling window of this app focused". So Ctrl+W while typing in the editor was consumed by the OS hotkey and closed the main window's terminal tab.

**Fix:** `registerGlobalShortcuts` now detects when a sibling window of the same app has focus (`getAllWebviewWindows()` + `isFocused()`) and unregisters all OS-level shortcuts in that case. Background operation (another app foreground) still keeps shortcuts registered via a light 2s polling re-check. The file-viewer window now closes itself on Ctrl+W/Cmd+W, and the native macOS menu handler ignores key equivalents while the main window lacks focus.

### State maintenance for the file-viewer window

Verified the viewer is a genuine Tauri `WebviewWindow` with its own webview/state, then added:

- **One editor per file** — reopening a file focuses the existing window (deterministic label from connection+path) instead of duplicating it with conflicting edit state.
- **Unsaved-changes guard** — closing a dirty editor (Ctrl+W, X button) prompts Save / Don't Save / Cancel; a failed save keeps the window open.
- **Restart persistence** — open editors are recorded via `editor-window-changed` events into `r-shell-open-editors` (localStorage) and reopened on next launch; the viewer polls session health and reloads the file once the backend session is restored.

## Files

- `src/lib/keyboard-shortcuts.ts` — sibling-window focus awareness for global shortcut registration
- `src/App.tsx` — `openEditorWindow` (reuse/focus), editor tracking listener, restart reopen
- `src/FileViewerWindow.tsx` — Ctrl+W close, session-health polling, close guard wiring
- `src/components/file-editor-view.tsx` — close-requested guard + Save/Discard/Cancel dialog, opened/closed event reporting
- `src/lib/editor-windows-store.ts` — persistence + deterministic window label (new)
- `src/locales/en.json`, `src/locales/zh-CN.json` — 3 new keys (parity check passes)
- Tests: 16 new (store, close guard, viewer session recovery, sibling-window shortcuts)

## Verification

- Full suite: 664/664 pass (62 files), including 16 new tests
- `tsc --noEmit` clean; eslint 0 errors; `pnpm i18n:check` passes


src/FileViewerWindow.tsx
```diff
@@ -21,13 +77,33 @@ export function FileViewerWindow() {
...
+  // registerGlobalShortcuts), so Ctrl+W / Cmd+W reaches this webview instead
+  // of closing the main window's active terminal tab. Here it closes this
+  // editor window, matching how an editor closes its focused tab. Unsaved
+  // changes are protected by FileEditorView's close guard.
+  useEffect(() => {
+    const onKeyDown = (e: KeyboardEvent) => {
+      if ((e.ctrlKey || e.metaKey) && e.key.toLowerCase() === "w") {
+        e.preventDefault();
+        e.stopPropagation();
+        void getCurrentWindow().close();
```
GOODBOY008: **[blocking]** This `close()` — and in fact every close path in the viewer — is denied by the ACL at runtime. `capabilities/file-viewer.json` grants only `core:default` + `positioner:default`, and `core:window:default` does **not** include `allow-close`/`allow-destroy` (checked against `src-tauri/gen/schemas/acl-manifests.json`: the default set is getters/queries + `internal-toggle-maximize` only).

There's a second, sneakier path with the same root cause: registering `onCloseRequested` (file-editor-view.tsx:215) makes tauri-core `prevent_close()` every native close attempt once a JS listener exists (`tauri` 2.11.2 `src/manager/window.rs:171`), and the API's own wrapper then calls `window.destroy()` when the handler doesn't `preventDefault()` (`@tauri-apps/api/window.js`, `onCloseRequested` → `this.destroy()`). So the **X button on a clean editor also fails** — Rust refuses the native close because a listener exists, JS's `destroy()` is denied → the window can't be closed at all.

Fix: add `core:window:allow-close` + `core:window:allow-destroy` to `capabilities/file-viewer.json`, and add `.catch` here so a future denial surfaces instead of becoming an unhandled rejection (unit tests mock these APIs, so nothing catches this class of bug).

src/App.tsx
```diff
@@ -1152,24 +1161,45 @@ function AppContent() {
...
+
+    // Reuse an already-open editor for this file: bring it to the front and
+    // restore it if minimized. Its content state is maintained in its own
+    // webview, so nothing needs reloading here.
+    try {
+      const windows = await getAllWebviewWindows();
+      const existing = windows.find((win) => win.label === label);
+      if (existing) {
+        try {
+          await existing.show();
```
GOODBOY008, line 1192: **[blocking]** Same ACL issue on the reuse path: `show`/`unminimize`/`setFocus` aren't in `core:window:default` and `capabilities/default.json` doesn't add them, so all three invokes are denied when the main window calls them. The `catch` swallows the denials and then we `return` — reopening a file whose viewer is minimized or behind other windows does nothing visible (silent no-op).

Fix: add `core:window:allow-show`, `core:window:allow-unminimize`, `core:window:allow-set-focus` to `capabilities/default.json`. Once added, please also log in that catch — as written it masks genuine failures as "already visible/focused".

src/FileViewerWindow.tsx
```diff
@@ -21,13 +77,33 @@ export function FileViewerWindow() {
...
   }, []);
 
+  // While this window has focus, the main window suppresses its OS-level
+  // shortcut registrations (sibling-window detection in
+  // registerGlobalShortcuts), so Ctrl+W / Cmd+W reaches this webview instead
+  // of closing the main window's active terminal tab. Here it closes this
+  // editor window, matching how an editor closes its focused tab. Unsaved
+  // changes are protected by FileEditorView's close guard.
+  useEffect(() => {
+    const onKeyDown = (e: KeyboardEvent) => {
```
GOODBOY008, line 80: **[high — macOS]** This DOM handler never fires for Cmd+W on macOS. The native menu binds `close_connection` ("Close Tab") to `CmdOrCtrl+W` (`src-tauri/src/lib.rs`), and menu key equivalents consume the event before the webview sees a keydown — that's exactly why `MACOS_NATIVE_MENU_ACCELERATORS` skips these on macOS. So Cmd+W with the viewer focused: menu fires → main window drops the event via the new `document.hasFocus()` guard (App.tsx) → **nothing happens**. Before this PR it closed the wrong tab; now it's a dead key, so the PR description's "closes itself on Ctrl+W/Cmd+W" holds only on Windows/Linux.

Suggestion: `listen('menu-action')` here too and `close()` when `payload === 'close_connection'` while `document.hasFocus()` — symmetric with the main window's guard, and it keeps the menu as the single macOS dispatch point.

src/FileViewerWindow.tsx
```diff
@@ -21,13 +77,33 @@ export function FileViewerWindow() {
     void moveWindow(Position.Center).catch(() => {});
```
GOODBOY008: **[medium]** The deterministic labels from `editor-windows-store` make this mount-effect re-center conflict with `tauri-plugin-window-state` (registered with default `StateFlags::all()`, `lib.rs`). With timestamped labels the plugin never had saved state to restore; now it does, and on reopen (same session or next launch) it restores the label's size **and position** — which this `moveWindow(Position.Center)` then throws away. Net behavior: the user's size persists but position never does, and it's inconsistent with the main window, whose position is deliberately restored (`restore_main_window_position`).

I'd drop this re-center: App.tsx already centers at creation time, and the plugin's own restore validates monitor intersection and falls back gracefully. Its original purpose (cover a skipped creation-time placement) is superseded by deterministic labels + plugin restore.

GOODBOY008 (commented):
Verified empirically in a clean worktree (fresh `pnpm install`, tests + tsc + lint + i18n against the actual Tauri/wry/tauri-plugin sources in the lockfile).

**What's solid**
- The sibling-window focus model is the right design: blur → immediate full-registration sync → async refinement, 2s poll only while blurred, dispose-safe, first-match-wins semantics preserved. Windows/Linux Ctrl+W is fully fixed by this.
- Single-writer persistence (viewers emit `opened`/`closed`, main window owns localStorage) correctly avoids assuming shared storage between webviews.
- The session-health poll is well-gated: `sessionWasMissing` means a remount only happens when the initial load actually failed, so unsaved edits are never clobbered by a remount; `get_session_health` returns camelCase (`sshConnected`) — verified against `SessionHealth`'s `rename_all`.
- Close-guard event accounting is exactly-once per path (`allowCloseRef` early-return prevents double `closed`).

**Verification**
- `pnpm test`: 653/653 pass across 61 files, incl. all 4 new files (the PR body's 664/62 doesn't quite reproduce — `connection.test.ts` is a live-SSH integration test that vitest doesn't execute).
- `tsc --noEmit` clean; `pnpm i18n:check` passes (3 new keys, `common.save` reused).
- `pnpm lint` reports 1 error, but in `settings-modal.tsx` which is byte-identical to main — pre-existing, not from this PR.

**Findings** — 2 blocking (capability ACL denials that make close/reuse no-op at runtime), 1 high (Cmd+W is a dead key in the viewer on macOS), 3 medium, details inline. The architecture is right; the blockers are small additive fixes to `capabilities/*.json` plus a `menu-action` listener.

src/App.tsx
```diff
@@ -1203,10 +1233,44 @@ function AppContent() {
...
-  }, [activeConnection, t]);
+    return () => { unlistenPromise.then(fn => fn()); };
+  }, []);
+
+  // Reopen file-editor windows that were open when the app last quit. The
+  // viewer windows poll session health, so they load their files as soon as
+  // the backend reconnects or show the error/retry view when the connection
+  // is gone.
+  useEffect(() => {
+    const persistedEditors = loadOpenEditors();
```
GOODBOY008, line 1294: **[medium]** This effect re-runs whenever `openEditorWindow`'s identity changes, and that callback depends on `t` — which gets a new identity on a language switch (or i18n resource reload). Each re-run does show/unminimize/setFocus for every tracked editor, so changing the language mid-session pulls all editor windows to the front and the last one steals focus. A `useRef(false)` run-once guard matches the stated intent ("reopen on next launch").

src/components/file-editor-view.tsx
```diff
@@ -149,6 +179,78 @@ export function FileEditorView({
...
+  }, [guardWindowClose, emitWindowEvent]);
+
+  // Window close guard: unsaved changes prompt to save/discard, and the
+  // window's open/closed state is reported for persistence and restore.
+  useEffect(() => {
+    if (!guardWindowClose) return;
+    let disposed = false;
+    let unlisten: (() => void) | undefined;
+
+    const handleCloseRequested = (event: { preventDefault: () => void }) => {
```
GOODBOY008, line 211: **[medium — data loss]** The guard covers per-window `CloseRequested` (X button, Ctrl+W), but app quit destroys windows without per-window close-requested events — so a dirty editor closes with no prompt. Combined with the new restart persistence it's worse than a plain crash: next launch the editor silently reopens showing the disk version, which reads as "my edits vanished" with no signal. VS Code prompts on quit too.

Options: (a) broadcast a quit-requested app event from a Rust `RunEvent::ExitRequested` handler and let dirty viewers prompt/flush (needs a timeout so a stuck viewer can't block exit forever), or (b) explicitly scope this PR with the limitation documented (maybe at least warn-before-quit in a follow-up issue). Worth deciding deliberately rather than shipping the silent variant.

src/lib/editor-windows-store.ts
```diff
@@ -0,0 +1,89 @@
...
+
+/** Event payload emitted by viewer windows to the main window. */
+export interface EditorWindowEventPayload {
+  event: 'opened' | 'closed';
+  connectionId: string;
+  filePath: string;
+  fileName: string;
+}
+
+export const EDITOR_WINDOW_CHANGED_EVENT = 'editor-window-changed';
```
GOODBOY008: nit: missing newline at end of file.

PR #125, opened 2026-09-01
fix(ssh): support connecting to hosts without a password

### LINKED ISSUE #122: Unable to connect hosts without password
v2.9.0

I have a rpi based network device that allow connection without password.
I've added it to r-shell but when I try to connect, instead of immediate connection I get the connection settings window, then when I click Save got the error:

_Connection Error
An unexpected error occured while connecting._

There is no problem connecting to this host from Moba X-Term for example.
BTW: Sometimes I just want to save host info without even saving username and password, so it should not be mandatory to provide this info.

sunxiaobin89:
## Summary

Fixes #122 — SSH hosts that allow passwordless login (Raspberry Pi style devices with `PermitEmptyPasswords yes`, or hosts accepting the SSH `none` authentication method) could not be connected at all: clicking Connect opened the edit dialog, the dialog refused to save a blank password with "Password is required", and failed attempts showed a generic "An unexpected error occurred while connecting." toast.

## Root cause

The frontend treated a blank password as "no credentials configured". All five connect entry points (connect, session restore, tab duplicate, tab reconnect, quick connect) gated on `authMethod === 'password' ? !!password : !!privateKeyPath`, and the dialog's validation rejected blank passwords outright. The Rust backend (russh `authenticate_password(username, "")`) actually supports empty-password auth with no changes.

A secondary issue: Tauri `invoke` rejects with a plain string, but every connection-failure catch displayed `error instanceof Error ? error.message : <generic text>` — so the backend's specific "Authentication failed. Please check your credentials..." message was always swallowed by the generic fallback.

## Changes

- `src/lib/connection-config.ts`: add `connectionHasCredentials()` — a password-auth connection always qualifies (a blank password is a valid credential); publickey auth still requires a key path; anonymous FTP needs nothing
- `src/App.tsx`: replace the 5 duplicated `hasCredentials` checks with the shared helper; show the real Tauri error string in the 4 connection-failure toasts
- `src/components/connection-dialog.tsx`: drop the "Password is required" validation; use the real error string in the connect-failure toast
- `src-tauri/src/ssh/mod.rs`: for blank passwords, try `authenticate_none` first (hosts requiring no credentials grant it), then fall back to the empty-password request (PermitEmptyPasswords accounts). Non-blank passwords are unchanged.

## Testing

- Frontend: 723 Vitest tests pass, including 7 new ones (helper unit tests + dialog regression for blank-password save/connect via both the Save-onSave and Connect paths)
- Rust: 170 unit tests pass; 2 new `#[ignore]` integration tests in `src-tauri/src/ssh/tests.rs` verified against a local Docker sshd (Alpine OpenSSH, `PermitEmptyPasswords yes`, empty-password user): blank password connects, wrong password still rejected
- Manual: verified in the built app against the same Docker host — blank-password connection connects directly from the sidebar, and auth failures now surface the backend's real message

### DIFF
--- a/src-tauri/docker/empty-password-sshd/Dockerfile
+++ b/src-tauri/docker/empty-password-sshd/Dockerfile
@@ -1,0 +1,28 @@
+# Fixture for the passwordless-host integration tests in src/ssh/tests.rs
+# (issue #122). An Alpine OpenSSH server with user 'pi' whose password is
+# EMPTY and PermitEmptyPasswords enabled — mirrors the Raspberry Pi style
+# devices users report connecting to.
+#
+# Build & run:
+#   docker build -t rshell-empty-pass-sshd src-tauri/docker/empty-password-sshd
+#   docker run -d --name rshell-sshd-empty -p 2222:22 rshell-empty-pass-sshd
+FROM alpine:3.19
+
+RUN apk add --no-cache openssh shadow
+
+# User 'pi' with an EMPTY (but not locked) password — mirrors `passwd -d pi`.
+RUN adduser -D -s /bin/sh pi \
+    && passwd -d pi \
+    && mkdir -p /home/pi/.ssh && chown -R pi:pi /home/pi
+
+RUN ssh-keygen -A
+
+RUN printf '%s\n' \
+    'PasswordAuthentication yes' \
+    'PermitEmptyPasswords yes' \
+    'PermitRootLogin no' \
+    'Subsystem sftp internal-sftp' \
+    >> /etc/ssh/sshd_config
+
+EXPOSE 22
+CMD ["/usr/sbin/sshd", "-D", "-e"]
--- a/src-tauri/src/ssh/mod.rs
+++ b/src-tauri/src/ssh/mod.rs
@@ -170,26 +170,25 @@ async fn authenticate_session(
 ) -> Result<()> {
     let authenticated = match method {
         AuthMethod::Password { password } => {
-            // A blank password means the host may require no credentials at
-            // all (it grants the SSH "none" method) or have an empty-password
-            // account (PermitEmptyPasswords). Try "none" first for blank
-            // passwords — non-blank passwords go straight to password auth.
-            let granted = if password.is_empty() {
-                session
+            // A blank password can mean two things: the host has an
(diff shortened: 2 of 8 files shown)


sunxiaobin89:
Thanks for the review — addressed in a8144a3. Per-item notes, including a correction where the review's premise didn't match the observed behavior:

**1. Absent vs blank password (`connectionHasCredentials`) — fixed.** A password-auth record with an *absent* password field (hand-edited/imported rows) now stays "unconfigured" and opens the dialog, since `buildSshConnectRequest` serializes the field to `null` which `ssh_connect` rejects. An intentionally blank string ("") still connects directly. Covered by new unit tests (`password: undefined` / `null` / `''`).

**2. Fixture not in the repo — fixed.** The empty-password sshd fixture is now committed at `src-tauri/docker/empty-password-sshd/Dockerfile`, and the test comments reference it. Endpoints are overridable via `RSHELL_EMPTY_PASS_HOST`/`RSHELL_EMPTY_PASS_PORT`, mirroring the existing `RSHELL_TEST_SSH_*` pattern.

**3. Branch coverage of the "none" auth paths — partially addressed, with a correction.** The review's premise was inverted: on the PermitEmptyPasswords fixture (empty-password account), OpenSSH *grants* the `none` method (`authenticate_none` returns `true` — verified), so the original test actually exercised the none-granted path; what lacked coverage was the *rejection → fallback* path. While verifying this I found something more important: I initially reworked the order to try `none` first and fall back to the blank password, then tested it against an explicit `AuthenticationMethods password` server — OpenSSH **disconnects** clients that send a method not in that list. So the PR's original `none`-first order would have broken blank-password connects on such hosts. The final order sends the **blank password request first** and only attempts `none` after an explicit rejection (the session stays alive across a USERAUTH_FAILURE). I also verified that an `AuthenticationMethods` + `PermitEmptyPasswords` combination is unsupported by OpenSSH itself (sshd disconnects the blank-password request regardless of client behavior), so there is no fixture to cover it — and no real-world passwordless host uses that combination. The remaining uncovered branch ("none" granted after a blank-password rejection) cannot be constructed with OpenSSH and is documented in the test notes.

GOODBOY008 (approved):
@sunxiaobin89 LGTM

PR #128, opened 2026-09-06
fix(settings): start with an empty workspace when startup reconnect is disabled

### LINKED ISSUE #126: Settings option for automatic reconnect at start
We have different operating modes, and sometimes we leave several SSH sessions open and close the application. 
After restarting the application, we want the tabs to be visible, but not necessarily for all connections to be immediately reconnected and refreshed – especially since this can take a long time for several tabs and slows down startup. 

An option to enable/disable automatic reconnection of open tabs after application startup would be useful.

GOODBOY008:
## Summary

Follow-up to #127 / #126: with **Reconnect Sessions on Startup** turned off, the previous session's tabs were still reopened as `pending` placeholders — only the automatic reconnect was skipped. This PR makes the setting behave as expected: **the previous tabs are not reopened at all** and the app starts with a fresh, empty workspace.

## Changes

- `terminal-group-context.tsx` — `initializeState()` now checks the setting before loading the persisted layout. When disabled it skips the load **and discards the stored layout**, so re-enabling the setting later cannot resurrect stale tabs whose reconnect list was already wiped by the empty workspace (the mount-time "save active connections" effect clears the list whenever the workspace is empty).
- `App.tsx` — the restore effect's disabled branch lost the "Sessions Not Reconnected" toast and deferred-marker population (nothing exists to defer anymore); it is now a guard that returns before starting any connection.
- Renamed `deferredRestoreTabIds` → `failedPendingTabIds` (`App.tsx`, `terminal-callbacks-context.tsx`, `terminal-tab-portals.tsx`): the marker now only means "a pending tab whose connect attempt failed — offer Connect again". That UX is preserved (still no `PtyTerminal` mount while a connect is in flight; Connect re-offered after a permanent failure), with the hint reworded to "The last connection attempt failed".
- Locales — setting description updated in `en.json` / `zh-CN.json`; removed the now-unreachable toast strings. `pnpm i18n:check` passes.

## Tests

The old `restore-skipped-when-disabled.test.tsx` (471 lines) was built around the removed "restored but pending" behavior, so it was split into two files:

- **`startup-restore-tabs.test.tsx`** — uses the *real* `TerminalGroupProvider` (seeded via the serializer) to exercise the layout gating end-to-end:
  - default (setting absent) still reconnects every saved connection;
  - setting off → no tabs rendered, no `ssh_connect`, no toasts, layout discarded, reconnect list cleared.
- **`pending-tab-reconnect.test.tsx`** — keeps the on-demand reconnect semantics (entered via the tab context menu, which is the path that still exists): startup starts no connections; pending tab reconnects on demand; no `PtyTerminal` mounted while the connect is in flight; Connect re-offered after a permanent auth failure (no dead terminal, no backoff retry); credentials-dialog flow reconnects the exact clicked tab.

## Verification

- `pnpm test`: 74 files, 742 tests passing
- `tsc --noEmit` clean; `eslint` on touched files: 0 errors (the one repo-wide error in `settings-modal.tsx:425` is pre-existing from #127 and untouched here)

Closes #126 (the remaining UX gap reported there)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -133,10 +133,10 @@ function AppContent() {
   const [isRestoring, setIsRestoring] = useState(false);
   const [restoringProgress, setRestoringProgress] = useState({ current: 0, total: 0 });
   const [currentRestoreTarget, setCurrentRestoreTarget] = useState<{ name: string; host?: string; username?: string } | null>(null);
-  // Tabs restored from the previous session whose automatic reconnect was
-  // skipped because "Reconnect sessions on startup" is disabled. They stay
-  // `pending` and show a Connect action until the user reconnects them.
-  const [deferredRestoreTabIds, setDeferredRestoreTabIds] = useState<ReadonlySet<string>>(() => new Set());
+  // Pending tabs whose latest connect attempt failed. They stay `pending` and
+  // show a Connect action instead of the "waiting" placeholder, because
+  // nothing is actually in flight for them anymore.
+  const [failedPendingTabIds, setFailedPendingTabIds] = useState<ReadonlySet<string>>(() => new Set());
   // The exact tab whose Connect/Reconnect opened the credentials dialog, so
   // handleSaveConnection reconnects that tab. Matching by connection id alone
   // misses primary tabs (no originalConnectionId) and is ambiguous for
@@ -352,25 +352,12 @@ function AppContent() {
 
       if (!isRestoreSessionsOnStartupEnabled()) {
         // The user opted out of automatic reconnect at startup (issue #126).
-        // TerminalGroupProvider already restored the tab layout with every tab
-        // in the `pending` state; leave them there and start no backend
-        // connection. The active-connections list is intentionally kept so
-        // the tabs are persisted again for the next launch.
-        const restoredTabIds = new Set(
-          Object.values(stateRef.current.groups).flatMap(g => g.tabs.map(t => t.id))
-        );
-        const deferred = new Set(
-          activeConnections
-            .map(conn => conn.connectionId)
-            .filter(id => restoredTabIds.has(id))
-        );
-        console.log(`Session restore skipped by setting: ${deferred.size} tab(s) left pending`);
-        if (deferred.size > 0) {
-          setDeferredRestoreTabIds(deferred);
-          toast.info(t('app.restoreDeferred'), {
-            description: t('app.restoreDeferredDesc', { count: deferred.size }),
-          });
-        }
+        // TerminalGroupProvider started from a fresh, empty workspace and
+        // discarded the previous layout, so there is nothing to reconnect
+        // into. This branch is a guard: it keeps a stale active-connections
+        // list (e.g. after an effect-order change) from ever initiating a
+        // connection when the setting is off.
+        console.log('Session restore skipped by setting');
         return;
       }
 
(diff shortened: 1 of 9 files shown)


twkrol:
Thanks for the follow-up — the failedPendingTabIds rename and the split tests read better than what I originally sent.

That said, the new behaviour is the opposite of what I was asking for in [#126](https://github.com/GOODBOY008/r-shell/issues/126). My original request was:

After restarting the application, we want the tabs to be visible, but not necessarily for all connections to be immediately reconnected and refreshed – especially since this can take a long time for several tabs and slows down startup.

Keeping the workspace is the part I actually need: I want to still see which hosts I was working on, just without the reconnect storm at startup. With #128 the tabs are gone entirely, so the setting no longer solves that problem for me — rebuilding the tab set by hand on every launch is more work than waiting out the reconnects.

There is also a side effect I think is worth reconsidering regardless of which behaviour you prefer. initializeState() now does:

if (!isRestoreSessionsOnStartupEnabled()) {
  localStorage.removeItem(STORAGE_KEY);
  return createDefaultState();
}
That runs on every startup while the setting is off, so the saved layout is destroyed rather than just skipped. Someone who turns the setting off once loses their tab list permanently, and turning it back on cannot bring it back. If "start empty" is the behaviour you want, skipping the load without removing the stored layout would achieve the same thing without discarding user data.

Would you consider one of these?

Restore the previous behaviour — tabs reopened as pending, no automatic connection; or
Make it a three-way setting: reconnect (current default) / restore without connecting / start with an empty workspace. That would cover both your use case and mine.
Happy to implement whichever you prefer and open a PR.

GOODBOY008:
@twkrol Thanks for the detailed follow-up — and you're right, #128 went further than what #126 actually asked for.

Here's the direction I'd prefer, modelled on how Chrome handles session restore: when the app is quit and reopened, the previous workspace comes back, but only the tab(s) that were active at quit time are actually reconnected right away. The remaining tabs are restored in the background — visible as pending, reconnecting lazily without blocking startup or firing a reconnect storm at every host at once. On top of that, we'd remove the full-screen restore-progress overlay so the restoration happens quietly behind the UI rather than covering it.

As for the side effect you spotted — that part is intentional. When the setting is off, the app deliberately starts from a clean slate, so the stored layout is discarded rather than just skipped; otherwise the old tabs would resurface the next time the setting is re-enabled. We'd rather keep that behaviour as-is, because the Chrome-style restore above targets the actual use case from #126 — seeing which hosts you were working on without paying for a reconnect storm — so with that in place there shouldn't be a need to turn the setting off at all.

Nothing is set in stone, though — I'd really like to hear your thoughts on this direction: does it cover your use case, and how would you expect the background restore to behave? Any feedback or ideas are very welcome. And if you'd like to take on the implementation, a PR would be great — happy to review. Thanks again for pushing back on this.

twkrol:
That direction covers my use case exactly — seeing which hosts I was working on without every one of them being dialled at startup is the whole thing I was after. Happy to drop my "restore without connecting" idea in favour of it.

On the discard: fair enough, your call. If the Chrome-style restore lands, I agree the setting stops being something people reach for.

## How I'd expect the background restore to behave

On demand rather than eagerly: a restored tab connects the first time it is actually activated, not on a timer. That is what Chrome does with background tabs, and it is the only variant that removes the reconnect storm rather than just spreading it out — hosts you never open during a session are never touched at all. Concretely:

- at startup, connect the active tab only; every other restored tab stays `pending`;
- activating a `pending` tab starts its connect, showing the existing "Waiting for connection…" placeholder;
- a failed connect falls back to what `failedPendingTabIds` already does today — the tab offers Connect again.

The eager variant (connect everything in the background, sequenced) would still hit every saved host on every launch. For someone with a dozen tabs across slow or VPN-gated hosts that is the same cost, just moved off the critical path.

## A few details worth settling first

1. **Which tab counts as "active" with split groups?** Every group in the grid is visible at once, so "the tab that was active at quit" could mean the active tab of the active group only, or the active tab of each group. I lean towards each visible group's active tab — those panes are on screen, so a pending placeholder in them is exactly what the user is looking at.
2. **Does the same laziness apply to SFTP/FTP and RDP/VNC tabs?** They have their own connect paths; I would treat them identically and connect on activation.
3. **The setting's name.** Under this model "Reconnect Sessions on Startup" no longer describes what it does — it controls whether the workspace is restored at all. Something like "Restore Session on Startup", with the description covering the lazy reconnect, would match the behaviour better. Happy to leave it alone if you would rather not churn the locales again.
4. **Restore overlay.** Agreed it should go. With a single tab connecting there is nothing meaningful to show, and the tab's own placeholder covers it. Do you want any residual indicator (status bar, say), or nothing at all?

PR #136, opened 2026-09-08
fix(terminal): let the tab close button receive its click under the drag handler

### LINKED ISSUE #135: Tab close (X) button no longer closes the tab since [#133] - the click only selects the tab
Title: Tab close (X) button no longer closes the tab since #133 — the click only selects the tab

## Environment

- Build of `main` at e754358 (includes #133), Windows 11
- Not reproducible on v2.9.1 (3682bcf), which predates #133

## Steps to reproduce

1. Open two or more terminal tabs.
2. Hover a tab and click its X button.

## Expected

The tab closes.

## Actual

Nothing closes; the click only selects that tab. Closing still works from the tab context menu ("Close Tab" / "Close All Tabs") and via Ctrl+W.

## Cause

#133 added custom drag handling on the tab element: its `onPointerDown` calls `preventDefault()` and `setPointerCapture()` for every left press. The X button is a descendant of the tab, so a press on it bubbles into that handler and captures the pointer on the tab.

Pointer capture retargets `pointerup` to the tab, and the browser then dispatches the resulting `click` to the common ancestor of the pointerdown target (the button) and the pointerup target (the tab) — i.e. the tab itself. The button's `onClick` never fires; `handleTabSelect` runs instead. The context menu is unaffected because the handler returns early for non-left buttons.

The drag tests don't catch it because jsdom has no pointer capture.

## Fix

Stop `pointerdown` propagation on the close button so the drag handler never sees it — a one-line change plus a regression test that stubs `setPointerCapture` on the tab and asserts a press on the X button never reaches it. PR incoming.

twkrol:
Title: fix(terminal): let the tab close button receive its click under the drag handler

Fixes #135 

## Summary

Since #133 the tab element handles `pointerdown` for custom drag: it calls `preventDefault()` and `setPointerCapture()`. The close (X) button is a descendant, so a press on it bubbles into that handler and captures the pointer on the tab.

Pointer capture retargets `pointerup` to the tab, and the browser then dispatches the resulting `click` to the common ancestor of the pointerdown target (the button) and the pointerup target (the tab) — the tab itself. The button's `onClick` never fires; the press only selects the tab.

## Change

- `group-tab-bar.tsx` — `onPointerDown={(e) => e.stopPropagation()}` on the close button, so the drag handler never sees the press and the click reaches the button. The tab body keeps its drag behaviour; right-click and the context menu were never affected (the handler ignores non-left buttons).
- `group-tab-bar-drag.test.tsx` — regression test: stubs `setPointerCapture` on the tab (jsdom has none) and asserts a press on the X button never reaches it and the click dispatches `REMOVE_TAB`, while a press on the tab body still arms the drag.

## Verification

- Reproduced on `main` (e754358); v2.9.1 (3682bcf), which predates #133, is not affected.
- `pnpm vitest run` — 79 files, 769 tests passing (1 new).
- `pnpm tsc --noEmit` clean; `pnpm lint` — no errors in the changed file.

### DIFF
--- a/src/__tests__/group-tab-bar-drag.test.tsx
+++ b/src/__tests__/group-tab-bar-drag.test.tsx
@@ -90,6 +90,35 @@ describe('GroupTabBar custom drag', () => {
     expect(dispatch).not.toHaveBeenCalled();
   });
 
+  it('keeps a press on the close button out of the drag handler so the click closes the tab', () => {
+    // Regression: the tab's pointerdown handler calls setPointerCapture, which
+    // retargets pointerup to the tab; the browser then dispatches the click to
+    // the tab (the common ancestor) instead of the X button, so the close
+    // never fires and the press only selects the tab. jsdom has no pointer
+    // capture, so stub it on the tab and assert the press never reaches it.
+    const tabs = [makeTab('a'), makeTab('b')];
+    const { container } = render(<GroupTabBar groupId="1" tabs={tabs} activeTabId="a" />);
+    stubDropTarget(container, '1');
+
+    const tabA = getTabEl('a');
+    const capture = vi.fn();
+    (tabA as HTMLElement & { setPointerCapture: (id: number) => void }).setPointerCapture = capture;
+    const closeButton = tabA.querySelector('button') as HTMLButtonElement;
+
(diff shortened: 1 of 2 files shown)


GOODBOY008 (approved):
@twkrol Great catch! LGTM

PR #137, opened 2026-09-08
fix(terminal): keep a hidden pane's WebGL renderer through a grace period

### LINKED ISSUE #134: Significant delay switching tabs
v2.9.1
As the new version has many UI improvements, the tab-switching introduced a significant delay - sometimes more than a second to go from tab to tab (both has an active session)
I'm investigating, but you could take a look as well.

case 1:
fresh load, no sessions, opening 3 new connections, switching between them with DELAY

case 2:
session restore turned on, fresh load with 3 sessions loaded at startup, switching between them with NO DELAY

twkrol:
Fixes #134

## Summary

Since #105 a pane releases its WebGL context the moment it is hidden and rebuilds it on the next activation. On a large grid that makes every tab switch pay for tearing the renderer down and building it back up. This PR keeps the context alive for a grace period (60 s) after a pane is hidden, so quick switching costs nothing, while panes hidden for a long time still let go of the GPU — which is what #105 was for.

## What was measured

Instrumented fork builds (`[DIAG]` console timings on `main`), 248×45 grid, three SSH tabs, Windows 11 / WebView2. The same grid size, DOM size (~1.1k nodes), pane count and frame cadence (~270 fps idle) in every run — so the cost is not the size of the terminal or a throttled compositor.

| per switch | release on hide (current) | context kept |
|---|---|---|
| outgoing pane: `WebglAddon.dispose()` | 279–323 ms | 0.2–0.4 ms |
| incoming pane: activation (`focus()` waiting on the rebuilt renderer) | 300–328 ms | 5–6 ms |
| incoming pane: first completed paint after activation | +954–960 ms | +11–13 ms |
| **click → painted frame** | **~1300 ms** | **49–53 ms** |

Frame gaps around a slow switch were `343/939/9/9/4/…` ms: two discrete stalls (the dispose, then the rebuild + first paint), not a throttle. Creating the context itself is cheap (8–12 ms); it is the dispose/rebuild cycle on an already-used terminal that is expensive. Interestingly the same cycle costs ~5 ms on tabs restored at startup — I could not pin down that asymmetry, but the fix does not depend on it: the cycle is unnecessary work on every switch either way.

## Change

- `pty-terminal.tsx` — hiding a pane now arms a `HIDDEN_WEBGL_RELEASE_MS` (60 s) timer that releases the context; showing the pane again cancels it. Unmount clears a pending timer before disposing. Context loss while hidden is still covered by the existing `onContextLoss` fallback and the `ensure()` on activation, so the overnight-eviction case from #105 keeps working.
- `pty-terminal-renderer-lifecycle.test.tsx` — the existing hide test now expects no release right after hiding and a release once 60 s elapse; a new test covers hide/show inside the grace period (no dispose, no second `WebglAddon`, activation still repaints and focuses, the cancelled release never fires).

## Verification

- `pnpm vitest run` — 79 files, 769 tests passing (1 new).
- `pnpm tsc --noEmit` clean; `pnpm lint` — no errors in the changed file.
- Manual: a fork build of this branch on Windows 11 — tab switching with three freshly opened SSH tabs is immediate; the slow case from #134 no longer reproduces.

### DIFF
--- a/src/__tests__/pty-terminal-renderer-lifecycle.test.tsx
+++ b/src/__tests__/pty-terminal-renderer-lifecycle.test.tsx
@@ -295,7 +295,7 @@ describe('PtyTerminal renderer lifecycle (issue #87)', () => {
     expect(mocks.terminals[0].refresh).toHaveBeenCalledWith(0, mocks.terminals[0].rows - 1);
     expect(mocks.terminals[0].focus).toHaveBeenCalled();
 
-    // Hidden again: the GPU context is released.
+    // Hidden again: the GPU context is kept through the grace period …
     view.rerender(
       <PtyTerminal
         connectionId="lazy-1"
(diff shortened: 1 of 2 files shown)


GOODBOY008:
@twkrol I test terminal switch perfomance, It's really fast! Good job~

twkrol:
> @twkrol I test terminal switch perfomance, It's really fast! Good job~

Thanks. I would like to use it daily myself :)

GOODBOY008 (approved):

PR #140, opened 2026-09-08
fix(security): authenticate the local WebSocket PTY bridge

### LINKED ISSUE #138: [Security] Local WebSocket PTY bridge has no authentication or Origin check — any local process or web page can open/hijack SSH sessions
## Summary

The PTY bridge in `src-tauri/src/websocket_server.rs` listens on `127.0.0.1:9001–9010` and accepts every connection with no token and no `Origin` check. Any client that connects can drive the tagged `WsMessage` protocol against an already-authenticated SSH session. This includes `StartPty` (open a fresh interactive shell on the session), `Input` (inject keystrokes/commands into a running session), and the desktop `DesktopKeyEvent` / `DesktopPointerEvent` messages — all keyed only by a `connection_id` string.

Because browsers allow a WebSocket handshake to `ws://127.0.0.1` from any origin (this is not covered by CORS), **a malicious web page open in any browser** — not just a local process — can scan the ten ports and speak this protocol. The only thing standing between an attacker and a shell on the user's server is knowing/guessing the `connection_id`.

## Why the id is not sufficient protection

`connection_id` values come from two shapes:
- saved connections: `crypto.randomUUID()` (unguessable), and
- the quick/primary path: `` `connection-${Date.now()}` `` (`connection-dialog.tsx:387`), later suffixed `-dup-${Date.now()}`.

The second shape is two millisecond timestamps, a search space small enough for a persistent page to brute-force. Even for UUID ids, treating an unguessable string as an authentication secret is not a security boundary — the id is passed around the UI and is not designed to be secret.

## Impact

- A background process or a web page the user has open can send `StartPty { connection_id }` and receive a full interactive shell on the user's server, running as the user, with their SSH credentials already applied. PTY output is delivered only to the client that opened it, so a covert attacker gets their own shell rather than a view of the user's.
- `Input { connection_id, data }` lets an attacker inject commands into a session the user is actively using.
- This is an SSH client holding credentials to production hosts, so the blast radius is remote code execution on those hosts.

## Steps to reproduce

1. Connect to any SSH host in r-shell.
2. From another local program (or a web page in a browser on the same machine), open `ws://127.0.0.1:9001` (try 9001–9010).
3. Send a `StartPty` message with a plausible `connection_id`.
4. Observe a new PTY shell on the user's authenticated session; send `Input` to run commands.

## Suggested fix

- Generate a random per-launch token in the backend; hand it to the frontend via `get_websocket_port` (or a sibling command) and require it in the WebSocket handshake (subprotocol or first message). Reject connections without it.
- Reject handshakes whose `Origin` header is not the Tauri app origin (use `accept_hdr_async` to inspect headers).
- Use unguessable `connection_id` values everywhere (drop the `connection-${Date.now()}` shape in favour of `crypto.randomUUID()`).

Any one of the first two closes the hole; doing both is defence in depth.

## Environment

Found by code review of `main` @ e754358 (v2.9.1 line). Affects `websocket_server.rs` (bind/accept at lines 349–390, message dispatch from 544).

twkrol:
Fixes #138

## Summary

The PTY/desktop bridge on `127.0.0.1:9001–9010` accepted every connection. Loopback is reachable by every local process and — because WebSocket handshakes to `127.0.0.1` are not subject to CORS — by every web page the user has open. A foreign client that knew or guessed a `connection_id` could send `StartPty` / `Input` and obtain a shell on the user's already-authenticated SSH session. Quick-connect ids were `connection-<ms timestamp>`, which is brute-forceable.

This PR makes the handshake authenticated and the ids unguessable.

## Backend

- **Per-launch token.** `WEBSOCKET_TOKEN` (`lib.rs`) is a `OnceLock<String>` filled with 32 random bytes (base64url, no padding) right before the listener binds, so no client can connect before it exists.
- **Handshake validation** (`websocket_server.rs`). `accept_async` → `accept_hdr_async` with a callback that runs `validate_handshake(origin, query, expected)`:
  - the `token` query parameter must equal the launch token (constant-time compare; prefix/superstring/empty values fail; an uninitialised server token rejects everything);
  - if the client sends an `Origin` header it must be one of the app's own webview origins (`tauri://localhost`, `http(s)://tauri.localhost`); the Vite dev origin is accepted only in debug builds. Non-browser clients send no `Origin` and rely on the token alone.
  - Rejections answer `403` and are logged with the reason and origin.
- **`get_websocket_endpoint` command** returns `{ port, token }`. The token only ever crosses the webview ↔ backend IPC channel. `get_websocket_port` is kept as-is for compatibility.

## Frontend

- New `getWebSocketUrl()` helper (`src/lib/websocket-endpoint.ts`) used by both `PtyTerminal` and `DesktopViewer`; it puts the token in the query string (`ws://127.0.0.1:<port>/?token=<token>`). It falls back to the bare default port only when no backend answers (plain `pnpm dev` in a browser), where a connection cannot succeed anyway.
- `connection-dialog.tsx`: quick-connect ids are now `crypto.randomUUID()` instead of `connection-${Date.now()}` — nothing depended on the prefix.

## Tests

- Rust (`websocket_server.rs`, `handshake_tests`): allowed origins with a valid token; native client without `Origin`; foreign and `null` origins rejected even with a valid token; missing / wrong / prefix / superstring / empty token; token among other query parameters; uninitialised server token; generated token is 43 URL-safe chars and unique.
- Frontend (`websocket-endpoint.test.ts`): URL shape, percent-encoding of a non-URL-safe token, fallback when the backend is unavailable. Existing `PtyTerminal` test mocks switched to the new command.
- `pnpm vitest run` — 80 files, 771 tests passing (3 new). `pnpm tsc --noEmit` clean; `pnpm lint` — no errors in the changed files.
- `cargo test` — run on my fork's CI on macOS, Ubuntu and Windows (see the linked run); a Windows build of this branch connects and works normally.

## Notes for review

- Any one of token or Origin check closes the hole; both are applied as defence in depth.
- No change to the `WsMessage` protocol itself, so the frontend/backend wire format is untouched.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -2279,6 +2279,24 @@ pub async fn get_websocket_port() -> Result<u16, String> {
     }
 }
 
+/// What the webview needs to open the PTY bridge: the bound port and the
+/// per-launch token the server demands in the handshake (issue #138).
+#[derive(Debug, Serialize)]
+pub struct WebSocketEndpoint {
+    pub port: u16,
+    pub token: String,
+}
+
+#[tauri::command]
+pub async fn get_websocket_endpoint() -> Result<WebSocketEndpoint, String> {
+    let port = get_websocket_port().await?;
+    let token = crate::WEBSOCKET_TOKEN
+        .get()
+        .cloned()
+        .ok_or_else(|| "WebSocket server not yet started".to_string())?;
+    Ok(WebSocketEndpoint { port, token })
+}
+
 // ========== PTY Connection ==========
 // PTY terminal I/O now uses WebSocket instead of IPC for better performance
 // WebSocket server runs on a dynamically assigned port (9001-9010)
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -16,11 +16,18 @@ mod websocket_server;
 use connection_manager::ConnectionManager;
 use std::sync::atomic::AtomicU16;
 use std::sync::Arc;
+use std::sync::OnceLock;
 use tauri::{Emitter, Manager};
 use websocket_server::WebSocketServer;
 
 // Global atomic to store the WebSocket port (shared between backend and frontend)
 pub static WEBSOCKET_PORT: AtomicU16 = AtomicU16::new(0);
+
+/// Per-launch secret the PTY bridge requires in every WebSocket handshake.
+/// Generated when the server starts; handed to the webview only through the
+/// `get_websocket_endpoint` command, so a foreign local process or a web page
+/// that can reach 127.0.0.1 cannot open or drive sessions (issue #138).
+pub static WEBSOCKET_TOKEN: OnceLock<String> = OnceLock::new();
 
 /// Applies the saved top-left position of the "main" window on startup.
 ///
@@ -448,6 +455,7 @@ pub fn run() {
             commands::detect_gpu,
             commands::get_gpu_stats,
             commands::get_websocket_port,
+            commands::get_websocket_endpoint,
             // Standalone SFTP/FTP commands
             commands::sftp_connect,
             commands::sftp_standalone_disconnect,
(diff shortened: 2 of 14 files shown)


GOODBOY008 (approved):
Reviewed during the 2026-09-09 batch: per-launch token via OnceLock initialized before the accept loop, origin allowlist, constant-time comparison, and fail-closed semantics all check out. All WS client call sites converted. Merging to land the local bridge auth half of the security work (fixes #138).

PR #141, opened 2026-09-08
fix(security): verify SSH host keys against known_hosts, with a "trust new key" dialog and a working Settings switch

### LINKED ISSUE #139: [Security] SSH host keys are never verified — `check_server_key` always returns `Ok(true)` (MITM)
## Summary

`src-tauri/src/ssh/mod.rs:156–161`:

```rust
async fn check_server_key(
    &mut self,
    _server_public_key: &key::PublicKey,
) -> Result<bool, Self::Error> {
    Ok(true) // In production, verify the server key
}
```

Every server key is accepted unconditionally, for direct connections and for the jump-host/tunnel path alike. There is no `known_hosts` equivalent, no trust-on-first-use prompt and no warning when a host's key changes. The stub dates from the first commit (6bc149b) and is listed in AGENTS.md as a known pitfall, but it has never been tracked as an issue.

The **Host Key Verification** switch in Settings → Security (`settings-modal.tsx:1074`) is only persisted to localStorage; nothing in the backend reads it, so the UI suggests a protection that does not exist.

## Impact

Anyone positioned between the client and the server (a hotspot, a compromised router, DNS spoofing, an ARP-level attacker on the LAN) can present their own SSH server and the client will connect without complaint:

- with password authentication the **password is sent to the attacker's server**, and the attacker can relay the session to the real host and read/modify everything in it;
- with public-key authentication the session itself can still be relayed and observed;
- because the app stores credentials and auto-reconnects saved sessions at startup, a single hostile network exposes every saved host at once, silently.

Host-key verification is the one SSH mechanism that defends against exactly this, which is why every mainstream client treats a changed key as a hard error.

## Suggested fix

`russh-keys` 0.44 — already a dependency — ships the helpers:

- `russh_keys::check_known_hosts(host, port, &pubkey) -> Result<bool, Error>` (reads `~/.ssh/known_hosts`; returns `Ok(false)` for an unknown host and `Err(KeyChanged { line })` for a mismatch),
- `russh_keys::learn_known_hosts(host, port, &pubkey)` to record a newly accepted key.

A minimal, OpenSSH-compatible implementation:

1. In `check_server_key`, call `check_known_hosts`.
2. `Ok(true)` → accept.
3. `Err(KeyChanged)` → reject the connection and surface a clear "host key has changed" error in the UI (no override button on the happy path).
4. `Ok(false)` → unknown host: either accept and `learn_known_hosts` (plain TOFU, matching how most GUI clients behave), or ask the user with the key fingerprint. Either is a large improvement over the current state; the prompt can come later.
5. Apply the same check to the tunnel/jump-host session.
6. Wire the existing **Host Key Verification** setting to this behaviour, or remove the switch.

Sharing `~/.ssh/known_hosts` with OpenSSH means users who already trust a host from the command line get no extra prompts.

## Environment

Found by code review of `main` @ e754358 (v2.9.1 line). Reproduction: point a DNS name or `/etc/hosts` entry for a saved connection at a different SSH server — r-shell connects and authenticates without any warning.

twkrol:
Fixes #139 

## Summary

`check_server_key` accepted every server key unconditionally — a stub since the first commit — for direct connections and the jump host alike, so anyone between the client and the server could impersonate the host, capture the password and relay the session unnoticed. The **Host Key Verification** switch in Settings was persisted but never read.

This PR verifies host keys the way every mainstream client does, makes the switch real, and makes a changed key a decision rather than a dead end.

## Behaviour

| Situation | Result |
|---|---|
| host recorded, key matches | connect |
| host not recorded | record the key (trust on first use), connect, log the fingerprint |
| host recorded, key differs | connection refused; the app shows a **Host key changed** dialog with host, new fingerprint and the recorded line. *Cancel* leaves `known_hosts` untouched; *Trust new key and connect* replaces the recorded entry and reconnects |
| Settings → Host Key Verification **off** | no verification (logged as a warning) |
| `known_hosts` unreadable / malformed, or no home directory | refuse (fail closed) |

The jump host is verified with the same policy as the target. Standalone **SFTP** sessions build their own SSH client and follow the same policy (they share `known_hosts` with the terminal); a changed key there refuses the connection with the same message plus a hint to open an SSH terminal to the host, which offers the dialog — the SFTP path has no dialog of its own in this PR.

## Implementation

**Backend (`ssh/mod.rs`, `commands.rs`)**
- `HostKeyPolicy { Strict, AcceptNew, Off }` on `SshConfig`, parsed from a new optional `host_key_policy` request field (`"strict"` default, `"accept-new"`, `"off"`); legacy callers that omit it get `Strict`.
- `Client` carries host, port, the `known_hosts` path, the policy and a `HostKeyReport` slot. `check_server_key` runs `verify_host_key()`; a `Strict` refusal is stored as a typed `HostKeyChanged { host, port, fingerprint, line, file }` error, carried through `anyhow`, and `ssh_connect` returns it in a new `host_key_changed` field of `SshConnectResponse`. Other refusals keep a plain message. `connect()` and `connect_via_ssh_tunnel()` surface the stored reason instead of russh's generic "unknown key".
- `AcceptNew` replaces the recorded key: `forget_known_host()` drops the host's plain-text entries (`host` or `[host]:port`, also inside comma-separated alias lists) and keeps everything else — comments, other hosts, and hashed `|1|…` lines, which cannot be matched without the salt; then the new key is learned.
- **Why not `russh_keys::check_known_hosts`:** on Windows it resolves to `~/ssh/known_hosts` (no dot), which OpenSSH for Windows does not use. `default_known_hosts_path()` is `~/.ssh/known_hosts` on every platform and the `_path` variants of the `russh_keys` helpers are used, so hosts already trusted from the command line need no new decision and keys learned here are visible to `ssh`.

**Frontend**
- `buildSshConnectRequest` sets `host_key_policy` from the Settings switch (`'off'` only when explicitly disabled).
- Every `ssh_connect` call (six in `App.tsx`, one in the connection dialog) goes through a new `sshConnect()` helper. When the backend reports `host_key_changed`, the helper asks the mounted `HostKeyChangedDialog` (via a DOM event, so no prop drilling across the seven paths) and, on confirmation, retries **once** with `accept-new`. Declining, or having no dialog mounted, returns the original refusal — there is no silent escalation.
- Dialog strings added to `en`, `zh-CN` and `pl`; `pnpm i18n:check` passes.

## Tests

Rust (`ssh/mod.rs` → `host_key_tests`, no server needed; keys generated in-test, `known_hosts` in a temp dir):
- first contact records and is then recognised, no duplicate lines; non-default port is a separate `[host]:port` entry; a changed key is rejected and not recorded;
- `accept-new` replaces the key and keeps other hosts' entries; `forget_known_host` precision (comment, `[host]:port`, alias list, hashed line, missing file);
- handler: known/changed keys with the reason set, `Off` never touches `known_hosts`, `AcceptNew` admits a changed key once, a strict refusal downcasts to `HostKeyChanged`, no home directory refuses everything.

Frontend (`ssh-connect.test.ts`): policy from Settings (default strict, explicit off, garbage → strict); `sshConnect` passes a success through, retries with `accept-new` after the user trusts the key, keeps the refusal on decline or when no dialog is mounted.

`pnpm vitest run` — 80 files, 776 tests passing. `tsc --noEmit` clean, `pnpm lint` — no errors in the changed files. `cargo test` green on macOS, Ubuntu and Windows on my fork's CI (https://github.com/twkrol/r-shell/actions/runs/34287916062); a Windows build of this branch connects normally to hosts already in `~/.ssh/known_hosts`, records new ones, and shows the dialog when a key is swapped.

## Notes for review

- AGENTS.md pitfall 5 rewritten to describe the new behaviour.
- The branch carries a few follow-up commits (rustfmt of `ssh/mod.rs`, which also touches one pre-existing unformatted line at the public-key auth error; SFTP policy plumbing; test fixture updates). Squashing on merge is fine.


GOODBOY008 (approved):
Reviewed during the 2026-09-09 batch: known_hosts parsing verified against russh-keys 0.44.0 source (hashed |1| entries, per-type KeyChanged, the Windows ~/ssh/ vs ~/.ssh/ resolution), the HostKeyChanged downcast is safe because create_connection propagates it unwrapped, and the trust-new-key dialog plus Settings switch work end-to-end. Merging as the host-key half of the security work (fixes #139). Known follow-ups noted in review: the 15s restore timeout can race a slow trust-dialog answer; forget_known_host rewrite is not atomic.

GOODBOY008 (approved):
Reviewed during the 2026-09-09 batch: known_hosts parsing verified against russh-keys 0.44.0 source (hashed |1| entries, per-type KeyChanged, the Windows ~/ssh/ vs ~/.ssh/ resolution), the HostKeyChanged downcast is safe because create_connection propagates it unwrapped, and the trust-new-key dialog plus Settings switch work end-to-end. Merging as the host-key half of the security work (fixes #139). Known follow-ups noted in review: the 15s restore timeout can race a slow trust-dialog answer; forget_known_host rewrite is not atomic.

PR #155, opened 2026-09-14
feat(dev): keep dev/e2e builds off the OS keychain and skip update auto-checks

GOODBOY008:
## Summary

Dev/e2e builds get two unattended-automation guarantees: the OS keychain is never touched, and update checks never fire on their own. Every `tauri dev` rebuild changes the ad-hoc code signature, so macOS treats it as a new app and re-prompts for keychain access — which stalls AI-agent-driven e2e runs (and invalidates each "Always Allow"). An unexpected "update available" dialog or release-manifest fetch mid-run would break automation the same way.

## Tauri best-practice conformance

Audited against the official Tauri v2 docs ("Calling Rust from the Frontend") and the tauri 2.11.2 source:

- **Dev detection**: the official `tauri::is_dev()` API (true when the `custom-protocol` feature that `tauri build` enables is absent — exactly `tauri dev`, and `cargo test`, builds).
- **App paths**: the dev key file lives under the canonical `app.path().app_data_dir()`, so the identifier comes from `tauri.conf.json` (verified path-identical to the previous resolution: tauri's desktop `app_data_dir()` is `dirs::data_dir().join(config.identifier)`).
- **Threading**: "Commands without the async keyword are executed on the main thread." All commands this PR touches that do I/O are `async fn` with injected `tauri::AppHandle` (run via `async_runtime::spawn`, not the main thread): `credential_seal`, `credential_open` (first-run keychain/file I/O), and `get_update_context` (spawns `sw_vers`, stats Caskroom dirs) — matching the pre-existing `updater_check`. Async commands can't take borrowed args; ours take only owned values.
- **IPC surface**: injected handles are not part of the JS `invoke` args, so the frontend is unchanged; app-defined commands need no capability/ACL entries.
- **Frontend**: `invoke` from `@tauri-apps/api/core` (v2 path); dev state flows from the backend command rather than a frontend heuristic, keeping the backend the single source of truth.
- No `tauri` `test` dev-dependency added (it breaks the Windows CI loader).

## Changes

- **Keychain bypass** (`src-tauri/src/commands.rs`) — `master_key()` branches on a new `keychain_disabled()` check (`tauri::is_dev()`, or `RSHELL_DISABLE_KEYCHAIN=1` as an explicit opt-out for e2e of packaged builds):
  - Dev/e2e path: 32-byte master key stored at `<app-data>/dev-master-key.dat`, with `<app-data>` resolved through Tauri's `app.path().app_data_dir()` (base64, `0600` on unix, race-safe `create_new` with read-the-winner fallback, stable across relaunches so sealed secrets keep decrypting).
  - Release path: unchanged — still the single `com.aiden.r-shell.dataprotection` keychain entry.
  - Seal/open logic extracted into `seal_with_key`/`open_with_key` helpers and covered by a new `credential_key_tests` module: file-key roundtrip/stability, create-race convergence, corrupt/short-key rejection, seal↔open roundtrip, and `dev_builds_bypass_the_keychain`, which pins the contract that dev-profile builds never touch the keychain.
- **Auto-update check suppressed in dev/e2e** (`src-tauri/src/commands.rs`, `src/components/update-checker.tsx`) — `get_update_context` now reports `autoCheckDisabled` (`tauri::is_dev()`, or `RSHELL_DISABLE_AUTO_UPDATE=1` for packaged builds); `UpdateChecker` skips its startup auto-check when the flag is set, so automation never fetches release manifests or pops an update dialog. The context is still fetched, and **manual checks from Settings / the command palette keep working everywhere**. Contract test `dev_builds_disable_the_update_auto_check` pins the Rust side.
- **`AGENTS.md`** — pitfall #14 documents both dev/e2e suppressions, the env overrides, and the one-time re-entry consequence.
- **`src/lib/credential-crypto.ts`**, **`src/lib/update-channel.ts`** — header/type docs updated.

## Testing

- [x] `cargo test` — 205 passed, 0 failed (12 pre-existing ignored docker/live e2e), incl. the two dev-mode contract tests
- [x] `update-checker.test.tsx` — 34 passed (2 new: auto-check skipped when `autoCheckDisabled`, manual check still allowed)
- [x] Frontend vitest for the credential-crypto consumers (6 files) — 32 passed
- [x] `tsc --noEmit` clean
- [x] Empirical dev-profile check: real `master_key()` created the actual key file (`0600`, base64) and roundtripped a seal/open without any keychain access
- [x] `keyring` confirmed to appear only inside the gated branch (repo-wide grep)

## Notes for reviewers

- Secrets sealed by a dev build **before** this change fail to decrypt exactly once (frontend logs it and leaves the field as-is); re-entering the credential re-seals under the file key. Production users are unaffected.
- No CI runs on PRs in this repo (pull_request trigger removed); tests were run locally as listed above.


GOODBOY008 (commented):
Independently re-verified the whole PR against the tauri 2.11.2 source and the live test suites — **verdict: merge as-is; no re-implementation needed.**

**Design conformance (verified, not taken from the description):**
- `tauri::is_dev()` in the actual tauri-2.11.2 crate is `pub const fn is_dev() -> bool { !cfg!(feature = "custom-protocol") }` — exactly `tauri dev`/`cargo test` builds, never any `tauri build` output (including `--debug`). This is strictly better than `cfg!(debug_assertions)`, which would miss `tauri dev --release` runs.
- Repo-wide grep: `keyring::` appears only inside the gated branch of `master_key()` — the bypass guarantee has exactly one choke point.
- `updater_check` is invoked only via `checkForUpdates`; the only automatic caller is `maybeAutoCheck`, which now checks `autoCheckDisabled`. Read the full `UpdateChecker` — no interval/focus re-check paths that could bypass the flag; the `checkSignal` manual path is intentionally ungated (manual checks keep working, as designed).
- Async conversion + injected `AppHandle` is IPC-compatible: frontend `invoke('credential_seal', { secret })` sites are unchanged, and the Rust side re-checks on every `updater_check` anyway.

**Tests (re-run locally, all green):**
- `cargo test` — 205 passed, 0 failed, 12 ignored (incl. both dev-mode contract tests)
- `update-checker.test.tsx` — 34 passed
- `src/lib/__tests__` suite — 59 passed; `tsc --noEmit` clean

One P3 nit inline (theoretical only; not merge-blocking): the create-race loser can read the file in the microseconds between the winner's `create_new` and its `writeln!` completing, and would surface a one-time "corrupt" error instead of the winner's key. Requires two app instances hitting the first-use window simultaneously. If that ever bothers agent e2e launching parallel instances, a bounded retry (e.g. ~10 × 20 ms) on the `AlreadyExists` fallback closes it.

src-tauri/src/commands.rs
```diff
@@ -3642,11 +3642,91 @@ use base64::engine::general_purpose::STANDARD as BASE64;
...
+
+    use std::io::Write;
+    let mut file = match std::fs::OpenOptions::new()
+        .write(true)
+        .create_new(true)
+        .open(path)
+    {
+        Ok(file) => file,
+        Err(err) if err.kind() == std::io::ErrorKind::AlreadyExists => {
+            return read_file_key(path);
```
GOODBOY008, line 3709: P3 (theoretical, not merge-blocking): the race loser can hit this fallback while the winner is between `create_new` and its `writeln!` finishing, so `read_file_key` sees an empty/half-written file and returns a one-time "corrupt" error instead of the winner's key. Takes two instances hitting the first-use window within the same ~ms, and the error is clean and retryable — but if parallel agent e2e launches ever trip it, a bounded retry here (e.g. 10 × 20 ms on read failure) closes the window.

PR #157, opened 2026-09-15
feat(update): one-click copy for the Homebrew upgrade command

sunxiaobin89:
## Summary

Homebrew-managed installs (v2.9.3 P0 gating) direct users to run `brew upgrade --cask r-shell` themselves, but the command existed only as prose: mixed into a sentence in the guidance toast, and as a bare `<code>` block in the settings banner. Users had to manually select text to copy it.

## Changes

- **Settings banner** (Advanced tab, Homebrew-managed state): the command block gains a copy icon button (via `@tauri-apps/plugin-clipboard-manager`, same pattern as other components) with a success toast.
- **Guidance toast** (manual "Check for Updates" on a managed install): reworked from a text-embedded command to a two-line layout — explanation line, then the command in its own mono code block with an inline icon copy button matching the settings banner. The description copy no longer embeds the command in prose.
- i18n: en/zh-CN/pl updated in sync (`homebrewManagedDesc` rewritten without the embedded command; new `copyCommand`/`commandCopied` keys). The command text is consumed from the single source `settings.updates.homebrewManaged.command` in all four places (banner render, banner copy, toast render, toast copy).

## Testing

- 4 new tests: banner copy success, toast copy success (description JSX rendered and the icon button exercised), clipboard-failure stays quiet (no false success toast) for both surfaces, and non-managed installs render no copy button. Mutation-verified via cold-read review.
- Full suite: 817/817 pass, `tsc --noEmit` clean, `pnpm lint` 0 errors, `pnpm i18n:check` parity clean.
- Manually verified on macOS with a Homebrew-managed install (real cask receipt): banner and toast both copy the exact command; non-managed install shows no banner/button.


GOODBOY008:
@sunxiaobin89 Would you help to resolve confilcts?

GOODBOY008 (approved):

